PR #139, opened 2015-01-23
Fixes and memleaks plumbing

pinepain:
- Fix segfault when NULL passed to AMQPChannel constructor (`new AMQPChannel(null);` will cause segmentation fault instead of fatal error)`. This may happens when undefined or unitialized variable passed instead of proper `AMQPConnection` object.
 - Plumb numerous memleaks introduced in previous PR (mainly, by me). Sorry for that.
 - Various formating and typo fixes, unused variable removal.

### DIFF
--- a/amqp.c
+++ b/amqp.c
@@ -471,15 +471,15 @@ zend_function_entry amqp_connection_class_functions[] = {
 	PHP_ME(amqp_connection_class, getTimeout, 	arginfo_amqp_connection_class_getTimeout,	ZEND_ACC_PUBLIC)
 	PHP_ME(amqp_connection_class, setTimeout, 	arginfo_amqp_connection_class_setTimeout,	ZEND_ACC_PUBLIC)
 
-    PHP_ME(amqp_connection_class, getReadTimeout, 	arginfo_amqp_connection_class_getReadTimeout,	ZEND_ACC_PUBLIC)
-    PHP_ME(amqp_connection_class, setReadTimeout, 	arginfo_amqp_connection_class_setReadTimeout,	ZEND_ACC_PUBLIC)
+	PHP_ME(amqp_connection_class, getReadTimeout, 	arginfo_amqp_connection_class_getReadTimeout,	ZEND_ACC_PUBLIC)
+	PHP_ME(amqp_connection_class, setReadTimeout, 	arginfo_amqp_connection_class_setReadTimeout,	ZEND_ACC_PUBLIC)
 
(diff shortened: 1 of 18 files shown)


amqp.c
```diff
@@ -471,15 +471,15 @@ zend_function_entry amqp_connection_class_functions[] = {
 	PHP_ME(amqp_connection_class, getTimeout, 	arginfo_amqp_connection_class_getTimeout,	ZEND_ACC_PUBLIC)
 	PHP_ME(amqp_connection_class, setTimeout, 	arginfo_amqp_connection_class_setTimeout,	ZEND_ACC_PUBLIC)
 
-    PHP_ME(amqp_connection_class, getReadTimeout, 	arginfo_amqp_connection_class_getReadTimeout,	ZEND_ACC_PUBLIC)
```
lstrojny: Tabs vs. spaces issue?

tests/amqpexchange_publish_with_properties_ignore_unsupported_header_values.phpt
```diff
@@ -17,7 +17,7 @@ $ex->declareExchange();
 $attrs = array(
     'headers' => array(
         'null'     => null,
-        'object'   => new DateTime(),
+        'object'   => new StdClass(),
```
lstrojny: stdClass

tests/amqpexchange_declare_existent.phpt
```diff
@@ -23,6 +23,7 @@ try {
     $ex->setName($exchangge_name);
     $ex->setType(AMQP_EX_TYPE_TOPIC);
     $ex->declareExchange();
+    echo 'exchange ', $ex->getName(), ' declared', PHP_EOL;
```
lstrojny: Not sure I get the change, shouldn’t there be a related change in the EXPECT section?

pinepain: No, the code above should fail. This doesn't affect test case success or failure, just to make output more verbose (in case of test failure).

amqp.c
pinepain: Yeah. There are numerous of such issues, i just mess all changes into one (it is not the best practice).

tests/amqpexchange_publish_with_properties_ignore_unsupported_header_values.phpt
pinepain: Fixed.

Issue #152, opened 2015-02-25
error: too few arguments to function call, expected 9, have 7


ryandesign:
Hello, I am the maintainer of rabbitmq-c and php-amqp in the MacPorts package management system. I updated rabbitmq-c from 0.5.2 to 0.6.0 yesterday, and now realized php-amqp doesn't build anymore with this version. I tested php-amqp 1.4.0, 1.6.0beta2, and the latest version in your repository. The error is:

```
amqp_exchange.c:493:2: error: too few arguments to function call, expected 9, have 7
        );
        ^
/opt/local/include/amqp_framing.h:796:1: note: 'amqp_exchange_declare' declared here
AMQP_PUBLIC_FUNCTION
^
/opt/local/include/amqp.h:117:3: note: expanded from macro 'AMQP_PUBLIC_FUNCTION'
  __attribute__ ((visibility ("default")))
  ^
1 error generated.
```

pinepain:
Hi. Which rabbitmq-c version do you have? Against what rabbitmq-c did you build php-amqp? What rabbitmq-c library loaded?

ryandesign:
rabbitmq-c version 0.6.0.

pinepain:
Confirmed. Library check is wrong, I will fix that soon.

lstrojny:
Closing

travisghansen:
Any chance we can get a tarball release (on php.net) with this in it?

PR #153, opened 2015-02-26
Fix building against latest rabbitmq-c versions

pinepain:
- Fix rabbitmq-c version comparison to detect available features (this should fix issue https://github.com/pdezwart/php-amqp/issues/152).
- Remove outdated rabbitmq-c versions from `.travis.yml` (keep master branch + 3 latest stable versions).


lstrojny:
Thanks!

PR #157, opened 2015-04-08
Removed alias for declare as declare is a directive in php

c-datculescu:
While trying to use the AMQP extension with Symfony2 DI, because proxies are used when declaring a service as lazy, we get the following error:

```HP Parse error:  syntax error, unexpected 'declare' (T_DECLARE), expecting identifier (T_STRING) in...```

The commit fixes the issue.

### DIFF
--- a/amqp.c
+++ b/amqp.c
@@ -511,8 +511,6 @@ zend_function_entry amqp_queue_class_functions[] = {
 	PHP_ME(amqp_queue_class, getChannel,		arginfo_amqp_queue_class_getChannel,		ZEND_ACC_PUBLIC)
 	PHP_ME(amqp_queue_class, getConnection,		arginfo_amqp_queue_class_getConnection,		ZEND_ACC_PUBLIC)
 
-	PHP_MALIAS(amqp_queue_class, declare, declareQueue, arginfo_amqp_queue_class_declareQueue,	ZEND_ACC_PUBLIC | ZEND_ACC_DEPRECATED)
-
 	{NULL, NULL, NULL}	/* Must be the last line in amqp_functions[] */
 };
 
@@ -540,8 +538,6 @@ zend_function_entry amqp_exchange_class_functions[] = {
 
 	PHP_ME(amqp_exchange_class, getChannel,		arginfo_amqp_exchange_class_getChannel,		ZEND_ACC_PUBLIC)
 	PHP_ME(amqp_exchange_class, getConnection,	arginfo_amqp_exchange_class_getConnection,	ZEND_ACC_PUBLIC)
-
-	PHP_MALIAS(amqp_exchange_class, declare, declareExchange, arginfo_amqp_exchange_class_declareExchange, ZEND_ACC_PUBLIC | ZEND_ACC_DEPRECATED)
 
 	{NULL, NULL, NULL}	/* Must be the last line in amqp_functions[] */
 };

### CODE REVIEW

**Comment by pinepain** [comment]
I wouldn't do that for BC reason. It is less likely that anyone use it, but I saw a legacy code that depends on that methods. That would be gone in next major release.

I know that [BC is for suckers](http://blog.ircmaxell.com/2013/06/backwards-compatibility-is-for-suckers.html) and that we made minor changes that maybe break some BC multiple times, but removing method at all inside minors is not the best idea.

**Comment by lstrojny** [comment]
We can do that with 2.0

**Comment by pinepain** [comment]
We can do it in 2.0, but I guess we are not ready to rename 1.6 to 2.0 while 2.0 should deliver some real improvements rather then minimal changes. I prefer to discuss what should be done for 2.0 out of this issue.

**Comment by c-datculescu** [comment]
Ok, i think it makes sense to not BC in 1.6 and to not make 1.6 to 2.0. In the mean time we will probably have to maintain our own fork.

Thanks,
Cristian.

**Comment by lstrojny** [comment]
OK, agreed. Thank you!

**Comment by pinepain** [comment]
@c-datculescu can you show us the minimal example to reproduce the issue? I can't get how it happens that DI component call `declare()` method on `AMQPQueue` and `AMQPExchange` and even more, how it happens from code.

Forking will lead to situation that you'll have to monitor upstream and port features and fixes back to your fork, so maybe patching will be more appropriate solution.

**Comment by vstepanyuk** [comment]
To reproduce error you need symfony2. In any bundle you can define service \AMQPExchange as lazy.
services.yml:
```yaml
amqp_exchange:
   class: AMQPExchange
   lazy: true
```

c-datculescu:
Any news on this? Should we provide more test cases?

pinepain:
Hi, I'm sorry but I didn't spend much time on this issue. For the first look, from [Symfony doc for DI, Lazy Services section](http://symfony.com/doc/current/components/dependency_injection/lazy_services.html):

> At this point the retrieved $service should be a virtual proxy with the same signature of the class representing the service.

So probably that proxy causes this error.

As noted above, I realize that using PHP keywork `declare` is bad practice, but we already have this method in legacy codebase. And it marked as deprecated and scheduled for removal in next major release.

At this point we have two evils: one is this issue with auto-generated class that mimic our deprecated signature, the second one is to break BC in minor release or increment major release version number just for two deprecated methods removal.

As to issue, using custom proxy without `declare` method might be a use case (or something like that, just wrap native class with any other).

As to breaking BC or incrementing major release number, I'm not a big fan of neither both of this solutions, but if we dive deeper into the issue, it is less likely that someone still using `AMQPQueue::declare()` method. It is even removed from stubs.

Though, code above works fine:

```php
$conn = new AMQPConnection();
$conn->connect();

$ch = new AMQPChannel($conn);

$q = new AMQPQueue($ch);

$q->declare();

var_dump($q->getName());
```
so we cant even get who use `AMQPQueue::declare()` and if someone suppress warnings (it is common with legacy codebase) he will suddenly get fatals.

At this point I guess we can keep in mind that those who use legacy deprecated methods will less likely upgrade extension to latest version. 

Though, I know that there are working production code which uses that methods and that they use more or less fresh php-amqp extension versions.

So I hope wrapping native php-amqp classes with custom proxies might be acceptable temporary solution. If it will not solve the problem, then we'll really have to remove that deprecated method, even in minor release.

//cc @lstrojny

c-datculescu:
Hello.

Thanks for the proposed solution, we will either give up using these as services or going with custom proxy. 

Great job with the extension and waiting impatiently for 2.0 release.

Issue #155, opened 2015-03-30
RabbitMQ's Direct reply-to


anton-shirikov:
I've been trying to implement RabbitMQ's [direct reply-to](https://www.rabbitmq.com/direct-reply-to.html) and there seems to be a problem.

#### First attempt:
```php
// Establish connection, create a channel & exchange for publishing...
$replyQueue = new AMQPQueue($channel);
$replyQueue->setName('amq.rabbitmq.reply-to');

echo 'Publishing...' . PHP_EOL;
$exchange->publish('ping', null, AMQP_NOPARAM, [
    'reply_to' => 'amq.rabbitmq.reply-to',
]);

echo 'Waiting for reply...' . PHP_EOL;
$replyQueue->consume(function (AMQPEnvelope $envelope) {
   echo $envelope->getBody() . PHP_EOL;
}, AMQP_AUTOACK);
```

With amqp 1.4.0 the producer hangs on consume and rabbitmq logs contain this error:
```
Channel error on connection <0.601.4> (172.17.0.105:48186 -> 172.17.0.103:5672, vhost: '/', user: 'admin'), channel 1:
{amqp_error,precondition_failed,"fast reply consumer does not exist",
            'basic.publish'}
```
With amqp 1.6.0beta2 producer recognizes the error and throws exception:
```
PHP Fatal error:  Uncaught exception 'AMQPQueueException' with message 'Server channel error: 406, message: PRECONDITION_FAILED - fast reply consumer does not exist' in /var/code/producer.php:26
```

#### Second attempt (kinda working):

Judging from this reply http://permalink.gmane.org/gmane.comp.networking.rabbitmq.general/36792
I need to first consume from `amq.rabbitmq.reply-to`  before i send a request.

I've tried to find a workaround with amqp-1.6.0beta2's consume from multiple queues and it is working. The problem is I have to declare an extra "useless" queue:

```php
// Establish connection, create a channel & exchange for publishing...
$queue = new AMQPQueue($channel);
$queue->setName('amq.rabbitmq.reply-to');
$queue->consume(null, AMQP_AUTOACK);

$queue2 = new AMQPQueue($channel);
$queue2->setName('temp-queue');
$queue2->declareQueue();

echo 'Publishing...' . PHP_EOL;
$exchange->publish('ping', null, AMQP_NOPARAM, [
    'reply_to' => 'amq.rabbitmq.reply-to',
]);

echo 'Waiting for reply...' . PHP_EOL;
[...]

PR #158, opened 2015-04-10
Add ability to re-attach consuming callback

pinepain:
**Changes:**
 - add ability to re-attach consuming callback in `AMQPQueue::consume()`; 
 - add `AMQP_JUST_CONSUME` constant for that;
 - cleanup queue consumer tag when `AMQPQueue::cancel()` called;
 - add `AMQPQueue::getConsumerTag()` method.

All changes BC compatible, doesn't break current API, has no known performance impacts and documented in stubs. 

As a demonstration, test files can be used.

**Fixes:**
 - RabbitMQ's Direct reply-to, issue https://github.com/pdezwart/php-amqp/issues/155.

**Description:**

The main problem that have in php-amqp for now - `basic.consume` request and actual messages reading is one method, which leads to situation where we can't easily separate that two parts - consuming request and actual consuming. This is a legacy part and we even can't break it (we stick to semantic versioning, so no BC in minor release allowed).

I decided to add `AMQP_JUST_CONSUME` constant which prevent `AMQPQueue::consume()` from sending `basic.consume` request. This adds some magic and I'm not happy from such solution, but at this time this is minimal effective change that required. It solves problem and doesn't add significant complexity to logic and codebase.

**Example**:

```php
$conn = new AMQPConnection();
$conn->connect();

$channel  = new AMQPChannel($conn);
$exchange = new AMQPExchange($channel);

$q_reply_to = new AMQPQueue($channel);
$q_reply_to->setName('amq.rabbitmq.reply-to');
$q_reply_to->consume(null, AMQP_AUTOACK);

// running request-response logic

$q_reply_to->consume(function (AMQPEnvelope $message, AMQPQueue $queue) {
    echo $message->getBody() . ': ' . $message->getRoutingKey() . PHP_EOL;

    echo 'Received on ', $queue->getName(), ' queue', PHP_EOL;

    return false;
}, AMQP_JUST_CONSUME);
```

When `AMQP_JUST_CONSUME` constant passed as flag or it ORed with other  flag constants, any other flags are ignored (it is logical, because we don't send anything to broker).


stubs/AMQPQueue.php
```diff
@@ -350,4 +362,14 @@ public function getChannel()
     public function getConnection()
     {
     }
+
+    /**
+     * Get latest consumer tag. If no consumer available or the latest on was canceled empty string will be returned.
```
lstrojny: Can we use null for the error case?

pinepain: Yeah, but it adds more lines and few extra checks for this, and `is_empty()` and in `if` cases empty string will acts the same as null. And for now in `var_dump()` output we use empty string if no consumer tag set. But if you think `null` is more reasonable return value I will make such trivial changes.

lstrojny: Let’s go with null then, yes

pinepain: I changed to `null` in `var_dump()` output and in `AMQPQueue::getConsumerTag()`.

lstrojny:
Can you squash and add a changelog entry?

pinepain:
Done.

lstrojny:
Thank you!

makasim:
Given this example [amqpqueue_consume_multiple.phpt](https://github.com/pdezwart/php-amqp/blob/master/tests/amqpqueue_consume_multiple.phpt) is it somehow possible to know the exact queue the message came from?

pinepain:
Nope.

makasim:
@pinepain there are possibilities to implement it. Could you please look at https://github.com/pdezwart/php-amqp/issues/281

Issue #170, opened 2015-07-16
getDeliveryMode always return 0


stof:
This makes the method unusable, as 0 is not a valid delivery mode.

See https://github.com/swarrot/swarrot/issues/103 for the place where the issue was found

pinepain:
I'll give it a look. Which php, php-amqp, rabbitmq-c and RabbitMQ version do you use?

pinepain:
Please, take a look at [amqpexchange_publish_with_properties.phpt#L125](https://github.com/pdezwart/php-amqp/blob/222c90f27100777fcf3388c5e9cb30998137961c/tests/amqpexchange_publish_with_properties.phpt#L126) test case whether `AMQPEnvelope::getDeliveryMode()` returns `1`.

Can you provide reproducible example for cases when `AMQPEnvelope::getDeliveryMode()` should return `1` but returns `0` instead?

pinepain:
Confirmed. `delivery mode` property should be 1 for nonpersistent and 2 for persistent messages.

pinepain:
@stof Sorry for delay. Can you join us on discussing PR #172 that attempts to fix this issue?

lstrojny:
1.6.0 stable has been released

PR #172, opened 2015-08-31
Fix invalid delivery mode returned by AMQPEnvelop::getDeliveryMode

pinepain:
This PR:

 - Fix issue #170: invalid delivery mode `0` returned by `AMQPEnvelop::getDeliveryMode()` when no `delivery_mode` message property set (that means that delivery mode is 1 - transient).
 - Add support for [new Travis container-based infrastructure](http://docs.travis-ci.com/user/migrating-from-legacy/).
 - Add [rabbitmq-c v0.7.0](https://github.com/alanxz/rabbitmq-c/releases/tag/v0.7.0) to build matrix.

This changes are BC compatible (though it is highly recommended to test integration before update to newer php-amqp version that will include this changes, as always, actually).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ php:
   - 5.6
   - 7.0
 
+sudo: false
+
 services:
   - rabbitmq
 
@@ -17,15 +19,15 @@ matrix:
   allow_failures:
     - php: 7.0
 
-install:
-  - sudo apt-get update
-  - sudo apt-get -qq --fix-missing install valgrind
-
-  - wget -qO - http://www.rabbitmq.com/rabbitmq-signing-key-public.asc | sudo apt-key add -
-  - sudo add-apt-repository 'deb http://www.rabbitmq.com/debian/ testing main'
-  - sudo apt-get update
-  - sudo apt-get install --only-upgrade -y rabbitmq-server
-  - sudo service rabbitmq-server restart
+#install:
+#  - sudo apt-get update
+#  - sudo apt-get -qq --fix-missing install valgrind
+#
+#  - wget -qO - http://www.rabbitmq.com/rabbitmq-signing-key-public.asc | sudo apt-key add -
+#  - sudo add-apt-repository 'deb http://www.rabbitmq.com/debian/ testing main'
+#  - sudo apt-get update
+#  - sudo apt-get install --only-upgrade -y rabbitmq-server
+#  - sudo service rabbitmq-server restart
 
 env:
   global:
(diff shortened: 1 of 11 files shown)


lstrojny:
Can you check the build?

pinepain:
Yeah, working on it, something strange goes on - in my branch it is green while PR build red.

pinepain:
Done

pinepain:
@lstrojny What I did is interpreted missed `delivery_mode` as transient (1) which is mostly correct for RabbitMQ but incorrect for ActiveMQ. While this should work perfect for RabbitMQ and doesn't break BC, compatibility with other MQ brokers may will be lost.

I also experimented with ignoring `delivery_mode` property when it is `NULL`, which may somehow break BC (`AMQPEnvelop::getDeliveryMode()` will also return `null`, when delivery mode is unknown) - same like in Java API - [AMQP.BasicProperties.getDeliveryMode()](http://www.rabbitmq.com/releases/rabbitmq-java-client/v3.5.4/rabbitmq-java-client-javadoc-3.5.4/com/rabbitmq/client/AMQP.BasicProperties.html#getDeliveryMode%28%29).

So until we figure out which way is the most appropriate, I mark this PR as WIP.

stof:
@pinepain in my application, all messages are delivered as persistent (2) and I was still missing the delivery mode. So the interpretation of missing delivery mode looks weird.

stof:
ah, these messages were not persistent. Sorry for the wrong report.

pinepain:
The way with returning `NULL` for values that was not originally set looks more logical for me, so I would even propose to do that for all getters, not only for delivery mode. While it will be less confusing, it will be also consistent with Java API, which looks for me well-designed. The only part I really worry is BC.

stof:
it would be great if RabbitMQ was sending the delivery_mode flag even when it matches the default, as it seems that the default value is not part of the AMQP protocol...

@pinepain currently, it returns an invalid value in such case, which breaks the dispatching of a message reusing the delivery mode, and does not match any constant. So it looks fine to me (please make sure that passing ``null`` as delivery mode when submitting is supported, letting the server choose the default delivery mode, to be consistent)

pinepain:
Returning `null` from `AMQPEnvelop::getDeliveryMode` will definitely break BC - strict checks will fail. 

Looking through ActiveMQ docs and specs gives me that it support only AMQP 1.0 so from 0.9.1 we have RabbitMQ, Apache Qpid and OpenAMQ (doesn't look active at all). While I mostly familiar with RabbitMQ, brief googling gives me that Qpid operates delivery mode on client side and looks like that OpenAMQ too. So my suggestion about breaking BC ruined and I guess we can stick with suggesting default `delivery_mode` as transient OR even setting `delivery_mode` as transient on sending if it is not set (but it looks ambiguous for me and we may work in heterogeneous environment where other clients in other languages doesn't send any delivery_mode explicitly).

pinepain:
The only functional change that affects issue #170 is https://github.com/pinepain/php-amqp/commit/614cab5b5e4c50a7ba836aceb22593572a806ecb#diff-9b81a369f4ea42ead436323d99882d90R298: 

```diff
@@ -295,6 +295,8 @@ void convert_amqp_envelope_to_zval(amqp_envelope_t *amqp_envelope, zval *envelop
 
 	if (p->_flags & AMQP_BASIC_DELIVERY_MODE_FLAG) {
 		AMQP_SET_LONG_PROPERTY(envelope->delivery_mode, p->delivery_mode);
+	} else {
+		AMQP_SET_LONG_PROPERTY(envelope->delivery_mode, AMQP_DELIVERY_NONPERSISTENT);
 	}
 
 	if (p->_flags & AMQP_BASIC_PRIORITY_FLAG) { 
```

pinepain:
@lstrojny I guess we can safely merge this changes

.travis.yml
```diff
@@ -17,15 +19,15 @@ matrix:
...
-  - sudo service rabbitmq-server restart
+#install:
+#  - sudo apt-get update
+#  - sudo apt-get -qq --fix-missing install valgrind
+#
+#  - wget -qO - http://www.rabbitmq.com/rabbitmq-signing-key-public.asc | sudo apt-key add -
+#  - sudo add-apt-repository 'deb http://www.rabbitmq.com/debian/ testing main'
+#  - sudo apt-get update
+#  - sudo apt-get install --only-upgrade -y rabbitmq-server
+#  - sudo service rabbitmq-server restart
```
stof: this should be removed rather than being commented. The history is already kept by git, and it is much better at it than comments

pinepain: True. The original intention is to keep it here to be able to switch to newer RabbitMQ server version during development (while Travis machines install RabbitMQ from the same source, it is often outdated), so I guess it should be kept for now.

pinepain:
@lstrojny can we merge this and release?

lstrojny:
For sure! @pinepain do you want to ship another beta?

pinepain:
Thanks! @lstrojny let's have extra one beta just in case we missed something. Stable should be stable.

lstrojny:
OK!

pinepain:
@lstrojny JFYI: you tagged beta4 but didn't released on PECL.

lstrojny:
@pinepain I know, I am having trouble with the PECL account. Can login to other php.net services but not there. Trying to get help.

Issue #167, opened 2015-06-17
Update PHP7 compatibility status


sshilko:
Please update wiki about whether its php7-ready
https://github.com/gophp7/gophp7-ext/wiki/extensions-catalog

jippi:
:+1:

jderusse:
:+1: PHP7 beta 1 is released

guilhermeblanco:
**Tests** YES
**Docs** NO
**Works on PHP5** YES
**Works on PHP7** NO
**Good on PHP 7** NO

Does not compile, does not work as it currently stands on PHP7. I started to do changes to turn it into compatible through this compare: https://github.com/pdezwart/php-amqp/compare/master...guilhermeblanco:phpng
It's still heavily WIP and not compiling... yet. =(

pinepain:
@guilhermeblanco Nice job! Can you clarify whether PHP-7 changes will be BC for PHP-5?

lstrojny:
@guilhermeblanco awesome!

guilhermeblanco:
@pinepain They cannot be... too many "catastrophic" changes to try BC (specially around string and hash APIs).
I also took advantage of this update to remove deprecated features on the extension.

pinepain:
@guilhermeblanco I've looked through changes you've made. Impressive. How do you see extension roadmap?

P.S.: we have Vagrant file and provisioner script in repo, please update them according to your dev environment so others can join you.

guilhermeblanco:
@pinepain there're so many changes required that I don't know how to start sometimes.
Usually I take the https://wiki.php.net/phpng-upgrading page and choose one of the issues, then I start updating everywhere. However, this extension seems to lack a lot of update in the last couple of years that scares me sometimes.
A good example is this method: https://github.com/pdezwart/php-amqp/blob/master/amqp_connection.c#L901
It's been quite a while (IIRC, since PHP 5.3) that it's not required to call `zend_object_store_get_object` (in PHP7 this is now `Z_OBJ_P`) because ZPP already does the cast to you. This method could be updated into a PHP 5.3+ compatible way like this:

```cpp
PHP_METHOD(amqp_connection_class, getPassword)
{
	amqp_connection_object *connection;

	if (zend_parse_method_parameters(ZEND_NUM_ARGS() TSRMLS_CC, getThis(), "O", &connection, amqp_connection_class_entry) == FAILURE) {
		return;
	}

	/* Copy the password to the amqp object */
[...]

PR #173, opened 2015-09-01
Phpng

hanikesn:
These changes allow the extension to be used with PHP7. They are not backward compatible with earlier versions of PHP. Some tests are failing as the methods have been removed, but the tests are still there. I still want to replace some instances of `char*` with `zend_string*`. I would be happy to get some feedback here.

```
Number of tests :  115               115
Tests skipped   :    0 (  0.0%) --------
Tests warned    :    0 (  0.0%) (  0.0%)
Tests failed    :   12 ( 10.4%) ( 10.4%)
Expected fail   :    0 (  0.0%) (  0.0%)
Tests passed    :  103 ( 89.6%) ( 89.6%)
```
```
FAILED TEST SUMMARY
---------------------------------------------------------------------
AMQPConnection constructor with amqp.timeout ini value set [tests/amqpconnection_construct_ini_timeout.phpt]
AMQPConnection constructor with both amqp.timeout and amqp.read_timeout ini values set [tests/amqpconnection_construct_ini_timeout_and_read_timeout.phpt]
AMQPConnection constructor with timeout parameter in creadentials [tests/amqpconnection_construct_with_connect_timeout.phpt]
AMQPConnection constructor with channel_max, frame_max and heartbeat limits [tests/amqpconnection_construct_with_limits.phpt]
AMQPConnection constructor with timeout parameter in creadentials [tests/amqpconnection_construct_with_timeout.phpt]
AMQPConnection constructor with both timeout and read_timeout parameters in creadentials [tests/amqpconnection_construct_with_timeout_and_read_timeout.phpt]
AMQPConnection setTimeout deprecated [tests/amqpconnection_setTimeout_deprecated.phpt]
AMQPConnection setTimeout float [tests/amqpconnection_setTimeout_float.phpt]
AMQPConnection setTimeout int [tests/amqpconnection_setTimeout_int.phpt]
AMQPConnection setTimeout out of range [tests/amqpconnection_setTimeout_out_of_range.phpt]
AMQPConnection setTimeout string [tests/amqpconnection_setTimeout_string.phpt]
Compare version in package.xml and module [tests/package-version.phpt]
```


tPl0ch:
:100:

pinepain:
Why did you wiped out timeouts?

amqp_channel.c
```diff
@@ -455,44 +421,37 @@ PHP_METHOD(amqp_channel_class, setPrefetchSize)
...
 set the number of prefetches */
 PHP_METHOD(amqp_channel_class, qos)
 {
-	zval *id;
-	amqp_channel_object *channel;
+	amqp_channel_object *channel = AMQP_CHANNEL_OBJ_P(getThis());
 	amqp_connection_object *connection;
-	long prefetch_size;
-	long prefetch_count;
+	zend_long prefetch_size;
```
pinepain: Prefetch size and prefetch in rabbitmq-c defined as `uint32_t prefetch_size, uint16_t prefetch_count`, so we probably should take care of them here.

php_amqp.h
```diff
@@ -298,8 +184,9 @@ extern zend_class_entry *amqp_exception_class_entry,
 		zend_throw_exception(amqp_connection_exception_class_entry, verify_connection_error_tmp, 0 TSRMLS_CC); \
 		return; \
 
+// Adjust for offset
 #define AMQP_VERIFY_CONNECTION(connection, error) \
-	if (!connection) { \
+	if (!connection || connection > -1000) { \
```
pinepain: What does this magic number mean?

pinepain:
When it will work and tests become green, it may be useful for some audience. 

Main concern is that at this time end-user API BC is broken and changes are not BC for PHP 5.x branch. In https://github.com/pdezwart/php-amqp/issues/167#issuecomment-123039503 I explained a bit my point of view, but in general - current design should be changed completely (both user API and internals).

Apart that, as mentioned above, fixing broken user API BC and letting it run in PHP 7 may be an optional ad-hoc solution for someone. I'm saying this because there was few questions about PHP 5.2 support last year and no questions about running it on 5.6 and so on. We may discuss whether it is good idea to maintain PHP 7 support as separate branch or keep in in someone's fork. You know, this extension doesn't have very active community, we event have no hands to update docs and write some examples.

/cc @lstrojny

amqp_channel.c
hanikesn: Yeah, some proper bounds checking is probably also in order.

php_amqp.h
hanikesn: That's a hack that needs to be fixed.

As cou can see in amqp_connection_object_fetch_object the amqp_connection_object* now is at an offset from the zend_object*. Because the struct changed (because zend_object don't have a fixed size anymore and need to be end the end of objects). That means that when zend_object* is null, the connection you get via amqp_connection_object_fetch_object is not null, but at 0-offset.

hanikesn:
> Why did you wiped out timeouts?

That wasn't on purpose. I guess @guilhermeblanco removed them. I will add them back in, as I don't like breaking BC.

> current design should be changed completely (both user API and internals)
The thing is that people want to run PHP7 sooner rather than later. And missing extensions don't help there. And PHP7 is a big change in itself, so if we break BC too by only releasing amqp2 adoption will be hurt.

IMHO the extension is not that bad in shape codewise, maybe some small refactorings here and there and things should be fine.

valgrind on travis is doing something funny, but when I will add the deprecated timeouts stuff back everything should be fine.

guilhermeblanco:
@hanikesn I removed because it was not only deprecated, it was replaced by read and write timeout config.

tPl0ch:
@pinepain TBH I don't see a reason to keep compatibility with PHP5 with this branch. There is no better time to break BC than now.

And I agree with @hanikesn about the scope of changes.
> current design should be changed completely (both user API and internals)

This would not be the right way to go, especially when you say that you don't have a lot of hands to help out. I think small and iterative drive-by-refactorings make much more sense in this context. And getting a not perfect, but tested solution out there early is a good thing.

Every single module that we can additionally provide when PHP7 will (hopefully) be released in November will add to the adoption rate.

Regards
Thomas

pinepain:
@hanikesn, @guilhermeblanco I guess even while `{get,set}Timeout` is deprecated, it's not a good time to break BC. When users will be migrating to PHP7 forcing them to rewrite code will not help them at all. If their codebase works with latest PHP 5.6 and doesn't depend on deprecated features, then it should run on PHP 7 without any problems (in most cases).

@tPl0ch my note about design changing is more general and I totally agree that working now is better than perfect but never. My intention is to keep codebase compilable and workable with both PHP 5 and PHP 7. I will try to put my hands and provide concept of what I mean in a next two days.

hanikesn:
On my machine:
```
Number of tests :  115               115
Tests skipped   :    0 (  0.0%) --------
Tests warned    :    0 (  0.0%) (  0.0%)
Tests failed    :    1 (  0.9%) (  0.9%)
Expected fail   :    0 (  0.0%) (  0.0%)
Tests leaked    :    0 (  0.0%) (  0.0%)
Tests passed    :  114 ( 99.1%) ( 99.1%)
```
```
FAILED TEST SUMMARY
---------------------------------------------------------------------
Compare version in package.xml and module [tests/package-version.phpt]
```

pinepain:
Sorry for delay. Here is my PR https://github.com/pdezwart/php-amqp/pull/181 to add PHP 7 support keeping BC for PHP 5 branch.

hanikesn:
Closed in favor of #181.

Issue #62, opened 2013-06-13
Implement RabbitMQ tutorials


gjuric:
Since the documentation is practically non-existent, it would be nice if somebody would implement [tutorials from the RabbitMQ site](http://www.rabbitmq.com/getstarted.html).

Maybe this would be less time consuming then trying to update the whole documentation and it would certainly be of much help to the newcomers.

Repo with tutorials for other languages and implementations is located at [rabbitmq/rabbitmq-tutorials](https://github.com/rabbitmq/rabbitmq-tutorials)

A-j-K:
#174  Please give feedback there on that PR. Thank you.

bweston92:
Updates? I seen there was a PR but still open.

dedalozzo:
The example are here:

https://github.com/rabbitmq/rabbitmq-tutorials/tree/master/php-amqp

pinepain:
@lstrojny as there are already RabbitMQ tuttorials and less likely we have any further plan to do anything on this issue, let's close it.

dedalozzo:
I strongly suggest to link them in the README.

lstrojny:
@dedalozzo care to open a PR for it?

Closing this for now

PR #174, opened 2015-09-08
Add tutorials as outlined on the official RabbitMQ site

A-j-K:
Hello,

This is regarding issue #62 

I have implemented tutorials 1 thru 3 and created this PR for review and critical feedback.

Once I manage to get this into upstream (and have therefore implemented any changes after feedback) I will complete the remaining 3 tutorials 4 thru 6.

regards,
--Andy

Edit: I have added 4, 5 and 6. Didn't see a point in waiting. I may "polish" them later this evening but at least they all work as expected.

### DIFF
--- a/tutorials/README.md
+++ b/tutorials/README.md
@@ -1,0 +1,26 @@
+
+# RabbitMQ C AMQP tutorials
+
+## Introduction
+
+This directory holds example tutorial code that implement the six tutorials 
+found at: 
+
+ - <https://www.rabbitmq.com/getstarted.html>
+
+#### Prereqs:
+- PHP5.6.x
+- librabbitmq 0.5.x
+- php-amqp pecl extension 1.4.0
+
+In order to actually execute these scripts you will need a RabbitMQ server.
+
+The easiest way (and the method used to test these as they were created) is to use
+the official RabbitMQ Docker image:
+
+ - <https://hub.docker.com/_/rabbitmq/>
+
+####Tested against docker docker pull rabbitmq:management
+- docker run -d --hostname my-rabbitmq -p 5672:5672 -p 8080:15672 rabbitmq:management
+
+This is a work in progress.
--- a/tutorials/tut-1-hello-world.php
+++ b/tutorials/tut-1-hello-world.php
@@ -1,29 +1,34 @@
 <?php
 
-if(!extension_loaded("amqp")) {
+if(!extension_loaded("amqp")) 
+{
     die("AMQP module not installed");
 }
 
-class TutorialConsumer {
-    private $_conn;
-    private $_chan;
-    private $_queue;
-    public function __construct() {
-        $this->_conn = new \AMQPConnection(['localhost', 5672, 'guest', 'guest']);
-        $this->_conn->connect();
-        $this->_chan = new \AMQPChannel($this->_conn);
-        $this->_chan->setPrefetchCount(1);
-        $this->_queue = new \AMQPQueue($this->_chan);
-        $this->_queue->setName("queue-hello-world");
-        $this->_queue->declareQueue();
+class TutorialConsumer 
+{
+    private $conn;
+    private $chan;
+    private $queue;
+    
+    public function __construct() 
+    {
+        $this->conn = new AMQPConnection(['localhost', 5672, 'guest', 'guest']);
+        $this->conn->connect();
+        $this->chan = new AMQPChannel($this->conn);
+        $this->chan->setPrefetchCount(1);
+        $this->queue = new AMQPQueue($this->chan);
+        $this->queue->setName("queue-hello-world");
+        $this->queue->declareQueue();
     }
     
-    public function consume() {
-        $this->_queue->consume(
-            function(\AMQPEnvelope $message, \AMQPQueue $queue) {
+    public function consume() 
+    {
+        $this->queue->consume(
+            function(AMQPEnvelope $message, AMQPQueue $queue) {
                 $queue->ack($message->getDeliveryTag());
                 if($message->getBody() == "QUIT") { 
-                    $this->_queue->delete();
+                    $this->queue->delete();
                     exit(0);                 
                 }
                 echo "From '". $queue->getName() . "': " . $message->getBody() . "\n";
@@ -32,42 +37,49 @@ class TutorialConsumer {
     }
 }
 
-class TutorialProducer {
-    private $_conn;
-    private $_chan;
-    private $_exch;
-    private $_queue;
-    public function __construct() {
-        $this->_conn = new \AMQPConnection(['localhost', 5672, 'guest', 'guest']);
-        $this->_conn->connect();
-        $this->_chan = new \AMQPChannel($this->_conn);
-        $this->_queue = new \AMQPQueue($this->_chan);
-        $this->_queue->setName("queue-hello-world");
-        $this->_queue->declareQueue();
-        $this->_exch = new \AMQPExchange($this->_chan);
-        $this->_exch->setName("exchange-hello-world");
-        $this->_exch->setType(AMQP_EX_TYPE_FANOUT);        
-        $this->_exch->declareExchange();    
-        $this->_queue->bind($this->_exch->getName()); 
-        $this->_exch->publish('Hello World!');
-        $this->_exch->publish('QUIT');
+class TutorialProducer 
+{
+    private $conn;
+    private $chan;
+    private $exch;
+    private $queue;
+    
+    public function __construct() 
+    {
+        $this->conn = new AMQPConnection(['localhost', 5672, 'guest', 'guest']);
+        $this->conn->connect();
+        $this->chan = new AMQPChannel($this->conn);
+        $this->queue = new AMQPQueue($this->chan);
+        $this->queue->setName("queue-hello-world");
+        $this->queue->declareQueue();
+        $this->exch = new AMQPExchange($this->chan);
+        $this->exch->setName("exchange-hello-world");
+        $this->exch->setType(AMQP_EX_TYPE_FANOUT);        
+        $this->exch->declareExchange();    
+        $this->queue->bind($this->exch->getName()); 
+        $this->exch->publish('Hello World!');
+        $this->exch->publish('QUIT');
     }
 }
 
 // We use PCNTL to create new processes to handle
 // different ends of the queuing systems.
-if(!extension_loaded("pcntl")) {
+if(!extension_loaded("pcntl")) 
+{
     die("PCNTL module not installed");
 }
 
 $pid = pcntl_fork();
-if($pid == -1) {
+if($pid == -1) 
+{
     die("Fork failed\n.");
-}
-else if($pid) {
+} 
+else if($pid) 
+{
     $consumer = new TutorialConsumer;
     $consumer->consume();
 }
-else { 
+else 
+{ 
    $producer = new TutorialProducer;
 }
--- a/tutorials/tut-1-hello-world.php
+++ b/tutorials/tut-1-hello-world.php
@@ -42,7 +42,7 @@ class TutorialProducer
 {
     private $connection;
     private $channel;
-    private $exch;
+    private $exchange;
     private $queue;
     
     public function __construct() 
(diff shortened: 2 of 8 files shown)


lstrojny:
Very cool! One question: can we go with PSR-2 coding style for the examples, mostly meaning three changes:

 - Don’t use _ for restricted properties
 - Put method and class declaration opening braces in the next line
 - Last but not least (and probably not even mandated by PSR-2): can we use "new AMQPConnection", not "new \AMQPConnection"?

Otherwise, if you squash the commits I am happy to merge.

tutorials/tut-1-hello-world.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+if(!extension_loaded("amqp")) 
+{
+    die("AMQP module not installed");
+}
+
+class TutorialConsumer 
+{
+    private $conn;
```
lstrojny: connection

tutorials/tut-1-hello-world.php
```diff
@@ -0,0 +1,85 @@
...
+    private $chan;
```
lstrojny: channel

tutorials/tut-2-work-queues.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+if(!extension_loaded("amqp")) 
+{
+    die("AMQP module not installed");
+}
+
+class TutorialNewTask {
+    private $conn;
```
lstrojny: connection, channel

lstrojny:
Two small style complaints, sorry I missed them last time:
 - Can you make the variable names more verbose? connection, channel, exchange and so on
 - Can you use a space after if, while, function and the like

If you could squash the commits then, we are ready to merge. Thanks again, the examples will serve as a great goto place for people looking to grasp typical messaging patterns.

stellartech:
Sorry, git rebase/squash is a mystery to me, subversion user here.

A-j-K:
Lets say that again with the correct Github account this time. Sorry, git rebase/squash is a mystery to me, too much time on subversion for me.

tutorials/README.md
```diff
@@ -0,0 +1,26 @@
...
+ - <https://www.rabbitmq.com/getstarted.html>
+
+#### Prereqs:
+- PHP5.6.x
+- librabbitmq 0.5.x
+- php-amqp pecl extension 1.4.0
+
+In order to actually execute these scripts you will need a RabbitMQ server.
+
+The easiest way (and the method used to test these as they were created) is to use
```
pinepain: I would avoid proclaiming that it is the easiest way. On OS X and Windows installing docker is not the same as on Linux. Apart that, we already ship Vagrant environment with provisioning script. See details in [readme file  under Development section](https://github.com/pdezwart/php-amqp#development).

tutorials/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Introduction
+
+This directory holds example tutorial code that implement the six tutorials 
+found at: 
+
+ - <https://www.rabbitmq.com/getstarted.html>
+
+#### Prereqs:
+- PHP5.6.x
```
pinepain: We officially support PHP > 5.3 (and I think we just drop 5.3 from tests and at this time you can even run php-amqp even on 5.3).

tutorials/tut-1-hello-world.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+// We use PCNTL to create new processes to handle
+// different ends of the queuing systems.
+if(!extension_loaded("pcntl")) {
+    die("PCNTL module not installed");
+}
+
+$pid = pcntl_fork();
```
pinepain: Personally I prefer to avoid this complexity and keep two separate files: one consumer and another producer, something like in [php-amqplib demo](https://github.com/videlalvaro/php-amqplib/tree/master/demo).

tutorials/README.md
```diff
@@ -0,0 +1,26 @@
...
+- librabbitmq 0.5.x
```
pinepain: 0.5.x is too old, maybe it worth mention that latest one required? And if you have php-amqp extension as a requirement below, then if one install it it will already have librabbitmq. So this requirement may not be obvious for newcomers.

lstrojny:
@A-j-K the tutorials look very good already, do you have time to finish the rest?

tutorials/tut-1-hello-world.php
stellartech: @lstrojny, I read this above and was a bit "meh, no, I'm not following the php-amqplib demo style" because I thought my effort was clearer and better, one tutorial == one php file. Couldn't be simpler imho. I kind of stopped reading the feedback at this point as to implement this style isn't a touch up to the code, it's a revamp.

stellartech:
@lstrojny all six are done. Just not to @pinepain liking. As noted above, I'm not refactoring them to look like php-amqplib demo because I believe 6 example files for 6 tutorials is easier to consume quickly then the pain of wading through 25 php files and stringing them all together.

pinepain:
@stellartech (or @A-j-K ?) It's not about liking, it's just different point of view. Look into RabbitMQ tutorials, say, https://www.rabbitmq.com/tutorials/tutorial-two-php.html, for me it is more logical to follow their style, so that it would be much less confusing for newbies to jump into AMQP complexity. I didn't mean copy php-amqplib style exactly. OK, give me some time to show with a PR what I mean.

tutorials/tut-1-hello-world.php
pinepain: Sorry if my feedback sounds rude to you. I don't mean that you have to follow some specific style.

A-j-K:
@stellartech == @A-j-K just different accounts, same endpoint :)

@pinepain, the issue for me is the usual one, we all only have so much time to have fun outside of work hours for our favourite little hobbies. Recently I had to give a talk at a PHP UG and RabbitMQ was a part of it. Since the UG was PHP I restricted all the code I wrote to PHP, obviously.

So, from the RMQ website tutorial I followed their getting started and it uses php-amqplib. Which I started with. But being a pecl author I knew there was a C implementation somewhere and went looking. Eventually finding it here (started at pecl.php.org). Anyway, I want to implement the demos I had already done using the php-amqplib library. But it wasn't so easy given just stubs. In order to work out some of I needed to read the C code also. So, since I was passing by here I thought I'd drop you some docs to make the next persons life a little easier than I had found it. Obvious way was to implement the standard 6 tutorials (but in the quickest way I could that would have made immediate sense to myself). And then drop them into a PR for you guys. Hence this PR. But having to go back and refactor into 2 files per tutorial (12 in total), yeah, it's not a huge job I guess. But it's a job none the less considering I have to put together my test environment again blah blah (the PR is only worth anything if they actually work).

And the 1.4.0 thing you mentioned. Yeah, I agree with you. However, you do have an issue here. I initially started trying to get 1.6.x working. However, on CentOS (yum) and Ubuntu (apt-get) the package managers give you an old librabbitmq-c library which your 1.6.x branch fails to compile against. I did eventually get 1.6.x going but that required me to build librabbitmq-c myself (and some other stuff it needed) before 1.6.x compiled. There were some other issues I had to solve also.

If you are trying to write some "get going" docs then you want it to work to "what comes out of the box" easily. Otherwise within 30mins your customers will have walked to php-amqplib and simply given up.

The "how to build" 1.6.x instructions I put together really really not what your average PHP developer or DevOps dept want to bleed over when an easy pure PHP drop-in exists. This really needs solving.

From my point of view a PECL implementation should always win over a pure PHP version and you are "in competition" with it. So if you want to move forward and make this module shine and be really attractive some effort on docs and build are needed.

I'd like to help more here but like yours and everyone else's, time is the problem, always is.

pinepain:
@A-j-K  I appreciate your notes, valid critics and efforts you put on this PR, I understand how hard is to find some free time. I'm not against you, we are on the same boat. And I totally agree that his PR brings value to the project and I see your point on forking and having one file per tutorial. What I'm trying to say that it doesn't looks like common way of AMQP usage. Simple drop-in code like in RabbitMQ tutorials is more common way to teach users. Normally, producers and consumers are separated: they have different business logic and it is common when there are more than one consumer and they run on different machine. With simple plain code (like in tutors below or in tests) it is just to copy-paste that code and start extending it. I really respect your efforts and my point may be wrong or biased, that's ok, we are all studying, if so - just point me where I wrong.

JFI, there are already [Tutorials for using RabbitMQ in various ways ](https://github.com/rabbitmq/rabbitmq-tutorials), and for [php-amqp extension too](https://github.com/rabbitmq/rabbitmq-tutorials/tree/master/php-amqp), and they strictly follow RabbitMQ tutorials (https://www.rabbitmq.com/getstarted.html). It is looks reasonable to point to tutors implementation and RabbitMQ tutors page to keep entropy down and focus on more important [stuff](https://www.youtube.com/watch?v=MvgN5gCuLac) :smile:. Anyway, give it a look and say what do you think about that tutorials.

> In order to work out some of I needed to read the C code also

We have tests written in PHP, so one can give it a look before diving into C code. Sure, it would be nice to implement some extra demos which go beyond tutorials and demonstrates some extra patterns (like [Direct reply-to](https://www.rabbitmq.com/direct-reply-to.html), implemented in [tests/bug_gh155_direct_reply_to.phpt](https://github.com/pdezwart/php-amqp/blob/master/tests/bug_gh155_direct_reply_to.phpt)). It is in my TODO list for a long time already.

**As to notes about packaging, playground and so on**, on Centos php-amqp extension comes from [remi repo](http://rpms.remirepo.net/). I guess we have to note that in readme too. As to Ubuntu, situation is not so good, I guess we have to create some PPA for that to let users access fresh extension version. But as always, time is a problem. And not every stay on default PHP version which comes with distro, so at this time installing it from pecl covers 99% use cases. The rest 1% is for cloning repo and building from master branch.

I see your point about playground environment. At this time we already have vagrant support (check [Development](https://github.com/pdezwart/php-amqp#development) note in readme), but pointing to some exact playground with all tools already set up in the top of readme and keeping it simple is a point. What about some vagrant box, like Rasmuses php7dev one? Maybe even base on his box so that users may try different PHP versions easily. But it is for vagrant only and based on Debian, while some users may get used to Ubuntu or Centos or other distro.

**As to php-amqplib library**, I don't think we really competitors. At least I don't look on this from this point. php-amqp extension has persistent connections support and has less overhead at runtime, while php-amqplib library fits AMQP model much better than php-amqp extension, it stick to async nature which is how and for AMQP protocol was designed for and has cleaner API. And it is more easy to install php library via composer rather than pecl extension, so it is natural that php-amqplib is more popular than php-amqp extension. There was reports when both extension and library was used in the single project, so I can't agree that this two project are competitors, at least at this time.

A-j-K:
I came back to this to try and sort it out. However, I had to rebuild my VM with php-amqp support.

Someone mentioned above my README.md was wrong with librabbitmq 0.5.x and php-amqp-1.40

So I decided to try following instructions. I compiled librabbitmq v0.6.0 as stated.

But when I then try to compile php-amqp 1.0.6, 1.0.7, 1.0.8 or 1.0.9 I get:-

    In file included from /usr/local/include/amqp.h:740:0, from /tmp/php-amqp-1.0.6/amqp_exchange.c:37:
    /usr/local/include/amqp_framing.h:798:11: note: expected 'amqp_boolean_t' but argument is of type 'amqp_table_t'
    AMQP_CALL amqp_exchange_declare(amqp_connection_state_t state, amqp_channel_t channel, amqp_bytes_t exchange, amqp_bytes_t type, amqp_boolean_t passive, amqp_boolean_t durable, amqp_boolean_t auto_delete, amqp_boolean_t internal, amqp_table_t arguments);
    ^
    /tmp/php-amqp-1.0.6/amqp_exchange.c:477:2: error: too few arguments to function 'amqp_exchange_declare'

This is why I ended up with teh README I had, it was the only combination I could get to compile. So I guess you guys need some better build instructions and what exactly you need to build php-amqp because from all I can see there are problems (I did try other versions of librabbitmq other than 0.6.0 but same problem persists).

pinepain:
Hi!

1.0.* versions look really outdated, check what we support and what is live by looking into [build matrix](https://travis-ci.org/pdezwart/php-amqp) or [travis config file](https://github.com/pdezwart/php-amqp/blob/master/.travis.yml)

pinepain:
@lstrojny as there are already RabbitMQ tuttorials and less likely we have any further plan to do anything on this issue, let's close it.

PR #179, opened 2015-09-18
honour --with-libdir option

remicollet:
For distro using "lib64"


lstrojny:
That makes a lot of sense, thank you! You were on a roll yesterday :)

PR #180, opened 2015-09-18
add downstream info (Fedora + EPEL)

remicollet:
As "installation" refer to "some systems has php-amqp extension", worth to give explicit link when exists, especially when the name is not "php-amqp" (perhaps, only for official repositories)

Notice: even it package name is php-pecl-amqp (per Fedora Guidelines, "yum/dnf install php-amqp" works)


lstrojny:
Thanks!

PR #181, opened 2015-10-08
Refactor and add PHP 7 support (with BC)

pinepain:
Introduced changes:
 - add PHP7 support.
 - add `AMQPEnvelope::hasHeader()`, `AMQPExchange::hasArgument()` and `AMQPQueue::hasArgument()` to check whether specific header exists.
 - completely move to `zend object` custom objects (`AMQPEnvelope`, `AMQPExchange` and `AMQPQueue`).
 - use zend object on custom objects for properties storing ( `AMQPConnection` and `AMQPChannel`).
 - fix not properly deleted connection resource.
 - fix not properly allocated and freed `amqp_table_t` arguments table memory.
 
This PR doesn't introduce BC-breaking changes but introduce few new methods.

This PR changes shouldn't have any significant performance impact.

--------------

Memcheck build are allowed to fail when run on PHP 7 while valgring >= 3.8.0 required, but travis has the old one out of the box (3.7.0, they use Ubuntu 12.04LTS). We don't install newer one by our own to stay on new container-based infrastructure - builds runs much faster but sudo is restricted.

Test are green on both PHP 7 and PHP 5, both built with debug and zts (using [Rasmus' php7dev box](https://github.com/rlerdorf/php7dev)). Tests runt with and without valgrind.


sshilko:
hurray

hanikesn:
Nice work! I hope I can take a look at it this week.

lstrojny:
How is that PR related to #173?

pinepain:
No relations.

This PR is another implementation of PHP7 support with BC to PHP5 + internal refactoring + minor features added and some bugs fixed. #173 PR also introduces PHP7 support and fix some bugs, but it doesn't keep BC.

pinepain:
Are there any activity on this PR? Any recommendations, notes, comments, code review, questions?

lstrojny:
@pinepain I like the limited scope of the PR very much in the sense that it takes care of PHP 7 and nothing more. Reviewing the code I three very much related questions:
 - Given that we introduce that many macros, what do you believe is the best way forward to get rid of them again? Specifically, what’s your gut feel towards supporting PHP 5 in the future?
 - Is what we see in this PR already the most the minimal set of macros that we need?
 - Are there more places where better, more "natural" upwards compatibility can be reached. E.g. we could typedef `zend_long` in PHP 5 as far as I can see and then use `zend_long` everywhere. Therefore getting rid of the indirection `PHP5to7_param_long_type_t`. What do you think?

php_amqp.h
```diff
@@ -184,6 +99,58 @@ extern zend_class_entry *amqp_exception_class_entry,
 	*amqp_queue_exception_class_entry;
 
 
+#if PHP_MAJOR_VERSION >= 7
```
lstrojny: That looks like a leftover

pinepain:
@lstrojny Thanks for taking a look on changest! As to limited scope, sad to say that it is not so limited - it does two things at a time: refactor internals and after that adds php7 support. Technically, it can be split into two PR, but as refactoring done as a preparation for easier php7 support both changesets are in single PR.

*Preface:*

Based on [PHP Supported Versions](http://php.net/supported-versions.php) page, active support of PHP 5.6 ends on 28 Aug 2016 (in 9 months as of now) and security updates ends on 28 Aug 2017 (in 1 year, 9 months as of now).

Also based on [PHP RFC: PHP 7.0 timeline](https://wiki.php.net/rfc/php7timeline) we expect to get GA release to the end of October, 2015.

We already know that PHP 7 adds a lot of new features and brings significant speedup and reduce memory usage. From userland code it mostly compatible with well-written code which was kept up to date with latest stable PHP version. So for those who invested into their codebase, moving to PHP7 shouldn't be impossible task. While it is hard to guess when vast majority of PHP commutation migrate to PHP 7, we may expect a good interest to it, large frameworks like Symfony, Laravel and Zend Framework 2 (at least their PHP7 builds are green) already support PHP 7. 

*As to questions:* 

> - Given that we introduce that many macros, what do you believe is the best way forward to get rid of them again? Specifically, what’s your gut feel towards supporting PHP 5 in the future?

When we decides to drop PHP5 branches support, we simply replace macros calls with their bodies and run unused variables cleanup and alternatively use `zend_string` for string arguments handling.

While we are not introducing new changes too often neither doing rewrites or refactoring PHP5to7_* macro and typedef family shouldn't make significant problems.  Alternative to these macros is keeping two separate branches for PHP 5 and 7, which is the least thing I want to do while we'll have to maintain them two branches instead of one. 

> - Is what we see in this PR already the most the minimal set of macros that we need?

We can reduce their number by embedding `#ifdef` sections into code for those macros which used once, but to keep code readable I prefer not to do that. So final answer is yes, this is a minimal required number of macros for php-amqp codebase to support both PHP5 and 7 branches.

> - Are there more places where better, more "natural" upwards compatibility can be reached. E.g. we could typedef zend_long in PHP 5 as far as I can see and then use zend_long everywhere. Therefore getting rid of the indirection PHP5to7_param_long_type_t. What do you think?

Yes, I initially did exactly what you suggest - defined `zend_long`. But for the sake of consistency in denoting features that are branch-specific I explicitly `#define`d of `typedef`ed them with  `PHP5to7_` prefix.

hanikesn:
I like this PR better than mine. I get's rid of this custom memory management thing, and provides support for both PHP 5 and 7. Tests with and without memcheck enabled are running fine on my machine and I found no obvious problems. So :+1: from me.

pinepain:
@hanikesn Thanks for reviewing!

tPl0ch:
:+1:

lstrojny:
@pinepain thanks for the reply, makes all perfect sense. Ready to merge when you are :+1:

pinepain:
@lstrojny I cleaned up package.xml releases history and files related to build. pecl packaging looks working, but please, give it a close look to prevent possible oops.

lstrojny:
@pinepain wonderful, thank you!

pinepain:
@lstrojny fixed broken build, ready for beta release (1.7.0beta1)?

lstrojny:
Wonderful, thank you!

lstrojny:
Will ship a beta release tomorrow or Thursday.

pkruithof:
Very nice, thanks for all the hard work guys! :shipit:

pinepain:
@lstrojny trivial: 1.7.0alpha1 on pecl has beta status, which is a bit ambiguous.

lstrojny:
@pinepain I am always getting something wrong with a release, am I. Not revoking it now, too small of an issue. Thanks for noticing!

pinepain:
@lstrojny that's ok, sure, no need to revoke.

tompedals:
Thank you for taking the time to do this! :tada:

KIVagant:
Excellent work.

PR #186, opened 2015-11-16
Update readme file with docs [skip ci]

pinepain:
This PR targets adding docs to readme file and doesn't affect extension functionality.

Please, review introduced changes and overall text.

I guess we may later split info from readme file into md docs.


lstrojny:
Thanks!

Issue #188, opened 2015-12-15
`setPort()` and/or `getPort()` is broken on AMQPConnection


pkruithof:
On 1.7.0-alpha1:

```php
$conn = new \AMQPConnection(
    [
        'host' => 'localhost',
        'port' => 1234,
        'login' => 'guest',
        'password' => 'guest',
        'vhost' => '/',
    ]
);

var_dump($conn->getPort()); // int(1234)
var_dump($conn->setPort(5678)); // bool(true)
var_dump($conn->getPort()); // int(1234)
```

So either `setPort` incorrectly returns `true` here, or `getPort` is broken some way, I think.

pkruithof:
Same for AMQPQueue:

```php
$queue = new \AMQPQueue($channel);
var_dump(AMQP_IFEMPTY); // int(256)
var_dump($queue->setFlags(AMQP_IFEMPTY)); // bool(true)
var_dump($queue->getFlags()); // int(0)
```

pinepain:
Bug with `AMQPConnection::setPort()` confirmed and should be fixed in #189.

As to `AMQPQueue` flags, from [php_amqp.h, lines 73-74](https://github.com/pdezwart/php-amqp/blob/b5f8ad6da275665dfb871e478a676969bda02d3c/php_amqp.h#L73-L74):\
> /* passive, durable, exclusive, auto-delete, no-wait (see https://www.rabbitmq.com/amqp-0-9-1-reference.html#queue.declare) */
> #define PHP_AMQP_QUEUE_FLAGS (AMQP_PASSIVE | AMQP_DURABLE | AMQP_EXCLUSIVE | AMQP_AUTODELETE)

`AMQP_IFEMPTY` and `AMQP_IFUNUSED` are not valid queue flags, though, they are valid flags for `AMQPQueue::delete()` action.

So behavior you see is expected. JFYI, by default, AMQP_AUTODELETE flag is set on new queues.

pkruithof:
Thanks for the quick fix!

PR #189, opened 2015-12-15
Fix AMQPConnection::setPort() writes to wrong property

pinepain:
Fixes https://github.com/pdezwart/php-amqp/issues/188.

This is simple copy-paste garbage cleanup after refactoring.

No BC breaking, no extra requirements.

### DIFF
--- a/amqp_connection.c
+++ b/amqp_connection.c
@@ -843,7 +843,7 @@ PHP_METHOD(amqp_connection_class, setPort)
 		return;
 	}
 
-	zend_update_property_long(this_ce, getThis(), ZEND_STRL("read_timeout"), port TSRMLS_CC);
+	zend_update_property_long(this_ce, getThis(), ZEND_STRL("port"), port TSRMLS_CC);
 
 	RETURN_TRUE;
 }
--- a/amqp_queue.c
+++ b/amqp_queue.c
@@ -125,6 +125,9 @@ PHP_METHOD(amqp_queue_class, setName)
 
 	/* Set the queue name */
 	zend_update_property_stringl(this_ce, getThis(), ZEND_STRL("name"), name, name_len TSRMLS_CC);
+
+	/* BC */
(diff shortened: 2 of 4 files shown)


pkruithof:
Got another inconsistency: `AMQPQueue::setName()` does not return `true`, all the other setters do.

pinepain:
@pkruithof, good catch. Fixed and squashed. But returning `true` in setters in most places is useless while we just update object property. I guess later we will get rid of it, but for now let's be consistent. Thanks.

pkruithof:
I caught it because I'm writing a library which uses this extension as a driver :wink: 

I agree it's practically useless when false is never returned (AFAIK), but it doesn't hurt and indicates the operation worked. I wouldn't change it (but I'm obviously biased since I would have to update the library in this case).

pinepain:
@pkruithof, I already add `return true` to `AMQPQueue::setName()` in this PR, so now it is consistent (at least in this part).

pkruithof:
I know, it was a reply on this statement:

> I guess later we will get rid of it, but for now let's be consistent.

lstrojny:
@pinepain I consider returning true in the setters a bug. Can we make them return null now already?

pinepain:
@lstrojny, for sure, but this will break BC. Currently this is a trivial fix to stay consistent, though, I totally agree that in our case setters should return nothing (`@return void` annotation in stubs).

pinepain:
@lstrojny, can we merge this as it fixes major issue (see https://github.com/pdezwart/php-amqp/pull/189/files#diff-40f7cdf3f96d4346a1418ce7a7af8e92L846 line) as we start receiving more bug reports related to it #188 and fresh one #192 and roll pecl release?

lstrojny:
@pinepain sure

pinepain:
Just a tiny minor note, new alpha2 marked as beta =).

PR #194, opened 2015-12-22
Fix stalled connection resource cleanup

pinepain:
This PR:

 - add zval type check when reading `channel` property during channel resource fetching which caused segfaults;
 - fix stalled connection cleanup which caused `Library error` and prevent new persistent connection creation;
 - fix API breaking due to missed `AMQPConnection::getHeartbeatInterval() and  AMQPConnection::getMaxFrameSize` methods when rabbitmq-c < 0.6.0 used;
 - show effective value for frame size and heartbeat value when connection established, otherwise provided value to constructor or default one used;
 - remove deleted source file from win build config;
 - add compiled librabbitmq version early check.

This changeset doesn't break BC but it extend existent API usage.

### DIFF
--- a/amqp_connection.c
+++ b/amqp_connection.c
@@ -1056,6 +1056,7 @@ PHP_METHOD(amqp_connection_class, getUsedChannels)
 Get max supported channels number per connection */
 PHP_METHOD(amqp_connection_class, getMaxChannels)
 {
+	PHP5to7_READ_PROP_RV_PARAM_DECL;
 	amqp_connection_object *connection;
 
 	PHP_AMQP_NOPARAMS();
@@ -1063,13 +1064,11 @@ PHP_METHOD(amqp_connection_class, getMaxChannels)
 	/* Get the connection object out of the store */
 	connection = PHP_AMQP_GET_CONNECTION(getThis());
 
-	if (!connection->connection_resource || !connection->connection_resource->is_connected) {
-		php_error_docref(NULL TSRMLS_CC, E_WARNING, "Connection is not connected.");
-
-		RETURN_NULL();
+	if (connection->connection_resource && connection->connection_resource->is_connected) {
+		RETURN_LONG(connection->connection_resource->max_slots);
 	}
 
-	RETURN_LONG(connection->connection_resource->max_slots);
+	PHP_AMQP_RETURN_THIS_PROP("channel_max");
 }
 /* }}} */
 
@@ -1078,6 +1077,7 @@ PHP_METHOD(amqp_connection_class, getMaxChannels)
 Get max supported frame size per connection in bytes */
 PHP_METHOD(amqp_connection_class, getMaxFrameSize)
 {
+	PHP5to7_READ_PROP_RV_PARAM_DECL;
 	amqp_connection_object *connection;
 
 	PHP_AMQP_NOPARAMS();
@@ -1085,13 +1085,11 @@ PHP_METHOD(amqp_connection_class, getMaxFrameSize)
 	/* Get the connection object out of the store */
 	connection = PHP_AMQP_GET_CONNECTION(getThis());
 
-	if (!connection->connection_resource || !connection->connection_resource->is_connected) {
-		php_error_docref(NULL TSRMLS_CC, E_WARNING, "Connection is not connected.");
-
-		RETURN_NULL();
+	if (connection->connection_resource && connection->connection_resource->is_connected) {
+		RETURN_LONG(amqp_get_frame_max(connection->connection_resource->connection_state));
 	}
 
-	RETURN_LONG(amqp_get_frame_max(connection->connection_resource->connection_state));
+	PHP_AMQP_RETURN_THIS_PROP("frame_max");
 }
 /* }}} */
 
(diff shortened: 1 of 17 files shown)


prolic:
I currently got "Library error: a socket error occurred" when using perisistent connection. So I hope this can get merged and released, soon.

pinepain:
ping @lstrojny could you have a look on a changeset? Is is good enough to be merged? It would be nice to have some stable release with this, at leas we may schedule rc.

amqp_connection.c
```diff
@@ -29,6 +29,7 @@
 #include "php.h"
 #include "php_ini.h"
 #include "ext/standard/info.h"
+#include "ext/standard/md5.h"
```
lstrojny: I can’t think of an attack scenario but I would still like to not use MD5 anymore ever

lstrojny:
Changeset looks good, props for introducing what might lead to a proper connection pool! What’s your opinion about the hash function?

amqp_connection.c
pinepain: Say, other ext which extract credentials from persistent storage. I agree about md5, idea and approach taken from interbase ext (http://lxr.php.net/xref/PHP_7_0/ext/interbase/interbase.c#921) and it serves as a fool-proof.

pinepain:
I even start worrying whether everything ok with you. Hope you are fine.

As to hash function, md5 is more a fool proof there and a way to get shorter ID string here. Do you think we need to stick to stronger hash function, say sha256 or even 512? Anyway, we store all credentials in AMQPConnection props (which I also don't like) so here it is more just about havin fixed-size ID.

hanikesn:
It could be a problem in a shared enviroment, when you could force a collision you could get a persistent connection from someone else, right? But as you don't know the other hash this would be close to impossible. If there's a simple option to use something stronger I would use that otherwise I think it's fine.

pinepain:
I'm fine to use any of available hash functions, say [any of sha](http://lxr.php.net/xref/PHP_7_0/ext/hash/php_hash_sha.h) or any other available in PHP.

pinepain:
@lstrojny As to connection pool, are you talking about persistent connections pool? I gave a brief look and that might be worth to follow [memcached way](http://php.net/manual/en/memcached.construct.php) to specify persistent connection id explicitly. For BC we MAY keep current approach to use connection string as ID. How about this way?

pinepain:
Just added early rabbitmq-c version check.

Rationale: I saw some reports ([1](http://stackoverflow.com/q/34872937), [2](http://stackoverflow.com/questions/9520914/installing-amqp-through-pecl) and it appears time to time in issues here) when users try to build php-amqp agains unsupported rabbitmq-c version which leads to fancy build logs which are not informative for those who are not familiar with internals of php-amqp of rabbitmq-c.

So now for rabbitmq-c < 0.5.1 (0.5.2 is minimal supported version):

    checking for librabbitmq version... 0.4.1
    configure: error: librabbitmq must be version 0.5.2 or greater, 0.4.1 version given instead

for rabbitmq-c < 0.6.0 (recommended version):

    checking for librabbitmq version... 0.5.1
    configure: WARNING: librabbitmq 0.6.0 or greater greater recommended, current version is 0.5.1

lstrojny:
@pinepain let’s go with SHA224 or SHA256 then. Adding the librabbitmq check is great, thanks!

pinepain:
@lstrojny The main technical reason to use any hash function in this case was to have fixed-size unique resource ID string to not to use `emalloc`/`efree`. Just note - it's not about security, it's just hashing, the fact that connection string will not be stored as-is is just side effect of hashing.

Made a tiny investigation about hashes: if we plan to add any of hash function, then we'll have to explicitly depend on `hash` PHP extension (`ZEND_MOD_REQUIRED("hash")`). Normally, it present in PHP, except turned off manually, so we may whether ignore those who don't have `hash` installed (highly unlikely, but what if) of code will be covered with conditional sections like ```#if defined(HAVE_HASH) ... #endif` which is the last thing I want to add to this extension.

lstrojny:
Depending on ext/hash is fine from my point of view and let’s go with SHA2 for hygiene reasons.

pinepain:
I really can't get about what hygiene we are talking about. Will look forward to revert that md5 things back. It was too early to introduce such changes.

pinepain:
@lstrojny I removed storing connection hash by any way. If some decision requires explanation it probably not good one. Idea with md5 was not good enough.

Squashed and rebased.

At this point we are able add ability to have multiple persistent connections by adding extra key to credentials, say, `persistent_id` which will identify single persistent connection among other persistent connections to the same server. What do you think?

lstrojny:
Cool, that’s way easier to understand, thank you. Looks good to me. Wanna squash?

pinepain:
I don't think squashing required, commits looks atomic from problems they solve point of view.

lstrojny:
OK, fair enough. I’ll merge

lstrojny:
I'll release an alpha3

pinepain:
Thanks!

pinepain:
@lstrojny just pinging you about
> I'll release an alpha3

PR #199, opened 2016-02-12
Minor fixes for windows

narkq:
Just as title says.

I'm not sure though if I did everything right.
Tested with librabbitmq 0.6.0 + PHP7 + VS14, thread safe x86 release build as shared library.
Builds and kinda works, some tests are failing though.

Please review.

### DIFF
--- a/amqp_connection_resource.c
+++ b/amqp_connection_resource.c
@@ -173,7 +173,7 @@ int php_amqp_set_resource_read_timeout(amqp_connection_resource *resource, doubl
 	if (timeout == 0.) {
 		read_timeout = 0;
 	} else {
-		read_timeout = (int) (max(connection->read_timeout * 1.e+3 - .5e+3, 1.));
+		read_timeout = (int) (max(timeout * 1.e+3 - .5e+3, 1.));
 	}
 #else
 	struct timeval read_timeout;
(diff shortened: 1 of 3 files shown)


amqp_connection_resource.c
```diff
@@ -173,7 +173,7 @@ int php_amqp_set_resource_read_timeout(amqp_connection_resource *resource, doubl
 	if (timeout == 0.) {
 		read_timeout = 0;
 	} else {
-		read_timeout = (int) (max(connection->read_timeout * 1.e+3 - .5e+3, 1.));
+		read_timeout = (int) (max(timeout * 1.e+3 - .5e+3, 1.));
```
lstrojny: Why do we need this?

pinepain: Looks like my typo. I'll give it a closer look later.

pinepain: @lstrojny yes, that was my error during some refactoring. As this happens under `#ifdef PHP_WIN32` everything was fine on non-win platforms. Thanks @narkq for catching.

narkq: There's a comment just above that block.

    In Windows, setsockopt with SO_RCVTIMEO sets actual timeout
    to a value that's 500ms greater than specified value.
    Also, it's not possible to set timeout to any value below 500ms.
    Zero timeout works like it should, however.

To be honest, I did not check yet if this behavior of winsock is the same for windows 8 / 8.1 / 10, but as of windows 7 that was true.

lstrojny:
Alright, thank you!

Issue #23, opened 2012-11-14
provide feedback on AMQP_MANDATORY, AMQP_IMMEDIATE


bkw:
We have a request offering to sponsor work on callbacks for cases where AMQP_MANDATORY AND AMQP_IMMEDIATE are not satisfiable by the broker. 

Also there is this: https://bugs.php.net/bug.php?id=62786

Any volunteers wanting to get in contact with the sponsor mail @pdezwart, @lstrojny or me.

PR #200, opened 2016-02-14
Document the known problems with AMQP_MANDATORY

olemartinorg:
Hi all! First, thanks for this great extension.

I spent a lot of time today trying to debug why i got a ```Library error: (unknown error)``` when I attempted to ```consume()``` from a queue after publishing a message with the ```AMQP_MANDATORY``` flag set. I referenced the built-in stubs in PhpStorm that only told me this flag was available, and shortly explained what it was supposed to do - so i assumed it was implemented and working fine in php-amqp.

After an extensive search, I found [bugs.php.net/62786](https://bugs.php.net/bug.php?id=62786) and #23. Since I'm guessing this won't be solved any time soon, I went ahead and documented the problems with this flag.

Please correct me if I did something wrong. If this gets merged, I'll send a pull-request to [JetBrains/phpstorm-stubs](https://github.com/JetBrains/phpstorm-stubs) so that they'll update their amqp.php stub.

### DIFF
--- a/stubs/AMQP.php
+++ b/stubs/AMQP.php
@@ -66,11 +66,16 @@ define('AMQP_IFUNUSED', 512);
 
 /**
  * When publishing a message, the message must be routed to a valid queue. If it is not, an error will be returned.
+ * Note: This flag does not currently work in pecl-amqp, and WILL cause undefined behavior if you use it.
+ * @link https://bugs.php.net/bug.php?id=62786
+ * @link https://github.com/pdezwart/php-amqp/issues/23
  */
 define('AMQP_MANDATORY', 1024);
 
 /**
- * When publishing a message, mark this message for immediate processing by the broker. (High priority message.)
+ * When publishing a message, the message must be routed to a valid queue with an available consumer. If it is not, an
+ * error will be thrown. If this flag is not set, the server will queue the message, but with no guarantee that it will
+ * ever be consumed.
  */
 define('AMQP_IMMEDIATE', 2048);
 
--- a/stubs/AMQPExchange.php
+++ b/stubs/AMQPExchange.php
@@ -160,7 +160,11 @@ class AMQPExchange
      * @param string  $routing_key The optional routing key to which to
      *                             publish to.
      * @param integer $flags       One or more of AMQP_MANDATORY and
-     *                             AMQP_IMMEDIATE.
+     *                             AMQP_IMMEDIATE. Note: AMQP_MANDATORY does
+     *                             not currently work and WILL cause undefined
+     *                             behavior if you use it. See
+     *                             https://bugs.php.net/bug.php?id=62786 and
+     *                             https://github.com/pdezwart/php-amqp/issues/23
      * @param array   $attributes  One of content_type, content_encoding,
      *                             message_id, user_id, app_id, delivery_mode,
      *                             priority, timestamp, expiration, type


stubs/AMQP.php
```diff
@@ -66,11 +66,16 @@
 
 /**
  * When publishing a message, the message must be routed to a valid queue. If it is not, an error will be returned.
+ * Note: This flag does not currently work in pecl-amqp, and WILL cause undefined behavior if you use it.
+ * @link https://bugs.php.net/bug.php?id=62786
```
pinepain: This is broker-specific (I guess in your case - RabbitMQ specific). Other AMQP 0.9.1 brokers may still use it.

stubs/AMQPExchange.php
```diff
@@ -160,7 +160,11 @@ public function getType()
      * @param string  $routing_key The optional routing key to which to
      *                             publish to.
      * @param integer $flags       One or more of AMQP_MANDATORY and
-     *                             AMQP_IMMEDIATE.
+     *                             AMQP_IMMEDIATE. Note: AMQP_MANDATORY does
+     *                             not currently work and WILL cause undefined
```
pinepain: Same here, `AMQP_MANDATORY` does work, but RabbitMQ ignores it.

stubs/AMQP.php
```diff
@@ -66,11 +66,16 @@
...
+ * @link https://github.com/pdezwart/php-amqp/issues/23
  */
 define('AMQP_MANDATORY', 1024);
 
 /**
- * When publishing a message, mark this message for immediate processing by the broker. (High priority message.)
+ * When publishing a message, the message must be routed to a valid queue with an available consumer. If it is not, an
```
pinepain: There is [AMQP spec](https://www.rabbitmq.com/resources/specs/amqp-xml-doc0-9-1.pdf) (section 1.8.3.7.5) and also [RabbitMQ doc](https://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.publish.immediate) which covers what this flag is for. If you use RabbitMQ, then check it [spec](https://www.rabbitmq.com/specification.html), it doesn't support all AMQP 0.9.1 features as-is, e.g. `immediate` flag is not supported. As RabbitMQ is not the only AMQP broker, I don't think that we should cover all broker-specific cases in stub comments.

pinepain:
Hi. I tried to provide comments to the diff, but it looks like more detailed explanation required.

From written I assume you are using RabbitMQ. And behavior you've seen is expected and covered by RabbitMQ docs, check [Compatibility and Conformance](https://www.rabbitmq.com/specification.html) page. And please, don't interpret stubs and comments in it as definitive manual. Stub purposes are code completion in IDEs and providing general comments, not detail docs, especially broker-specific.

I realize that this type of behavior you expected is weird, but after reading RabbitMQ AMQP protocol [compatibility](https://www.rabbitmq.com/specification.html) and [interoperability](https://www.rabbitmq.com/interoperability.html) things will be more clear for you.

Also note, while AMQP protocol has asynchronous nature, this extension uses synchronous approach, which leads to some ambiguity during debugging and exceptions analyzing. This is due historical reasons and underlying C library we use for AMQP protocol communication.

If you will have any questions - feels free to open an issue or you can also look through SO, there are plenty of  Q&A about AMQP, incl. in PHP & RabbitMQ integration. You can also try to ask question in [SO chat in PHP room](http://chat.stackoverflow.com/rooms/11/php), ping me there  as `@zaq178miami` if you'll require help with AMQP and PHP.

stubs/AMQP.php
olemartinorg: Yes, I know - I read the AMQP spec when writing that comment. IMO 'mark this message for immediate processing by the broker' doesn't really communicate the point that this message will fail only if no consumer is available to handle it immediately on that queue. I'm aware that RabbitMQ does not implement the immediate flag, but php-amqp will fail with an exception telling you so (which is why I didn't add any disclaimers on that flag).

stubs/AMQP.php
olemartinorg: Are you sure about that? RabbitMQ supports the MANDATORY flag (https://www.rabbitmq.com/specification.html) but if you look in #23 the problem seems to be that the underlying C library used by php-amqp uses a callback here, which doesn't work well with the synchronous way PHP (and thus php-amqp) is implemented. In other words, (although I haven't tried with other brokers) the problem here is in php-amqp, not RabbitMQ.

olemartinorg:
Hi there! Sorry, I read this comment after i replied to your inline code notes. I see that you're linking to some of the same resources I consulted before making this pull request.

Even if stubs are not the place for putting this, I don't really see the harm in them (as this is definitely an issue with php-amqp, and i could not find any other documentation explaining that unexpected behavior apart from the closed issue - #23). If this were in the stubs, it would have spared me half a day of debugging.

If the stubs are not a form of documentation, where is the documentation on the php-amqp API? I don't see the point in opening a new issue for this, as it has been reported before (you should rather re-open #23). The real issue is "the mandatory flag doesn't work", and this pull request was meant to be a "so let's at least document it until it gets fixed".

Again, thank you for your work! If any of my points come off as snarky, I promise you it is not intentional!

stubs/AMQP.php
pinepain: My bad. The problem is that php-amqp extension doesn't deal with `basic.return` method from broker. I slightly recall that #23 but mixed MANDATORY and IMMEDIATE from there, sorry for that. I'll add an issue (or better submit PR) for that.

pinepain:
Hi, good to know that we are both on the same wave.

First of all, The Docs. There are no fresh copy of them. Stubs and examples are probably the only docs for now, so I clearly see your point about adding more info to stubs. My point about stubs is that they are not **definitive** documentation. While I agree that documentation on flags MAY be improved, I'm against adding such info as bugs links. Bugs should be fixed, not [transformed to feature](http://www.catb.org/jargon/html/F/feature.html). Flags docs SHOULD be replaced with description taken from AMQP docs.

I will try to provide PR to add `basic.return` method support from broker, I'll ping you to test when it will be ready.

olemartinorg:
Would it work if I removed the links, then?

I agree that the docs for the flags should be copied more or less directly from the AMQP docs (but not verbatim, as "the server will return an undeliverable message with a Return method" does not make sense on the PHP-side of things, as that return message will be represented as something else, likely an exception). I don't know how others use the stubs (as documentation), but i looked in the stubs and examples to find out how php-amqp worked, and consulted the docs for my broker (and the AMQP spec) to figure out the details on what a specific function/flag.

If this is something you think will be fixed soon, there's no point in adding it to the docs/stubs - but if it drags out like #23, I'd vote to add it to the docs to warn others or make that ```Library error: (unknown error)``` a bit more descriptive. When I think about it, if you need to wait for upstream to fix something first, what about throwing an exception when the mandatory flag is used until a proper solution is in place?

pinepain:
Can you please provide minimal example to reproduce `Library error: (unknown error)` exception for your case?

olemartinorg:
Sure, i had some example code i was about to attach to an issue (before i found the previous issue and figured mine would just be marked a duplicate):

```php
$conn = new \AMQPConnection([
    'host' => 'host',
    'login' => 'username',
    'password' => 'password',
]);
$conn->connect();

$chan = new \AMQPChannel($conn);
$exchange = new \AMQPExchange($chan);

// Trouble:
$exchange->publish('test', '', AMQP_MANDATORY);

// No trouble:
//$exchange->publish('test', '', AMQP_NOPARAM);

$queue = new \AMQPQueue($chan);
$queue->setName('hello_world');
$queue->declareQueue();
$queue->consume(function(\AMQPEnvelope $message, \AMQPQueue $queue) {
    $queue->ack($message->getDeliveryTag());
    echo "Received message\n";
    return false;
});
```

I guess it should fail on the ```publish()``` call, but it instead fails on the ```consume()``` call. (I guess because of leftovers after the failed mandatory message.)

pinepain:
Thanks. I forgot to ask, what PHP, php-amqp, RabbitMQ, rabbitmq-c version do you use?

pinepain:
> I guess it should fail on the `publish()` call

Nope, `basic.publish` is async call. In case of success it returns nothing, while in case of failure it will return failed message with `basic.return` method from broker in case `mandatory` or `immediate` flag was set. In case no such flag was set, message silently dropped.

pinepain:
P.S.: 
> $queue->consume(function(\AMQPEnvelope $message, \AMQPQueue $queue) {

Note, second param to callback is always current queue callback called from and this param will be deprecated soon and removed later.

olemartinorg:
I'm using PHP 7.0.3, php-amqp 1.7.0alpha2, RabbitMQ 3.6.0-1 (the debian package from rabbitmq.com/debian) and librabbitmq-dev 0.5.2-2 (from the offical debian jessie repos, probably the only older package there).

> Nope, ```basic.publish``` is async call.

Ahh! That explains a whole lot.

> Note, second param to callback is always current queue callback called from and this param will be deprecated soon and removed later.

Yeah, i noticed that during testing - and i saw it in another issue. I found some workarounds, but looking forward to when it is removed.. :-) Go ahead and remove it whenever you like - my integration tests will probably catch it!

pinepain:
Hi, `basic.return` server method support available in PR #203.

This allows to handle properly returned messages which was published with `AMQP_MANDATORY` flag and couldn't be routed to target destination. 

Please, give it a look and test a bit. I would really appreciate any feedback you can provide.

lstrojny:
What do we want to do with this PR?

pinepain:
While PR #203 merged I guess this issue is not relevant anymore, I guess we can close it for now.

lstrojny:
Closing

PR #203, opened 2016-02-28
Add support for server method handling

pinepain:
This PR:
 - add basic.return method handling support, see [basic.return method description](http://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.return);
 - add publisher confirms support, see [Confirms (aka Publisher Acknowledgements)](https://www.rabbitmq.com/confirms.html).

Now it is possible to turn channel into confirms mode with `AMQPChannel::confirmSelect()` method and then be able to process `basic.ack` and `basic.nack` server methods (as well as `basic.return` methods when message was published as mandatory).

This changeset doesn't break BC but it extend existent API usage.

While there are added few new server methods handling support, when no callback set to handle that methods, PHP notice raised, which may potentially break some applications which manually convert notices to exceptions.


stubs/AMQPChannel.php
```diff
@@ -179,4 +179,69 @@ public function getConnection()
...
+    public function setConfirmCallback(callable $ack_callback=null, callable $nack_callback=null)
+    {
+    }
+
+    /**
+     * Wait until all messages published since the last call have been either ack'd or nack'd by the broker.
+     *
+     * Note, this method also catch all basic.return message from server.
+     */
+    public function waitForConfirm($timeout = 0.0)
```
lstrojny: This is seconds I guess? What about making the API microseconds based?

tests/amqpbasicproperties.phpt
```diff
@@ -0,0 +1,163 @@
+--TEST--
+AMQPBasicProperties
```
lstrojny: Isn't this just an AMQPMessage?

stubs/AMQPChannel.php
zaq178miami: Microseconds can be passed as a float part (after dot). Do we need to be microsec based? Overall precision is not good at all no, e.g. you can start waiting for 1.5 and it can be even 2.0 sec or smth, small values are not very precise, so I don't think we'll have any benefit from moving to microsec.

zaq178miami: But good note about guessing, I missed parameter docs.

tests/amqpbasicproperties.phpt
zaq178miami: This is amqp properties that for some reason was mixed with envelope (see it [in java](http://www.rabbitmq.com/releases/rabbitmq-java-client/v3.3.0/rabbitmq-java-client-javadoc-3.3.0/com/rabbitmq/client/Envelope.html)) fields. `basic.return` method holds propertis and no enveloper, so to not confuse userland, `AMQPBasicProperties` introducd and now `AMQPMessage` extends it. Later we'll get rid of `AMQPMessage` or at least make it looks more logical.

prolic:
I checked out your branch and tried to compile. Following error occurs:

`php-amqp/amqp_exchange.c:49:21: fatal error: com_err.h: Datei oder Verzeichnis nicht gefunden
compilation terminated.
Makefile:183: recipe for target 'amqp_exchange.lo' failed
make: *** [amqp_exchange.lo] Error 1`

Using rabbitmq-c 0.7.1 from http://github.com/alanxz/rabbitmq-c
PHP 5.6.16 on Ubuntu 15.10

prolic:
On master branch from https://github.com/pdezwart/php-amqp/ I can compile without problems.

zaq178miami:
@prolic Should be fixed now. My IDE auto-inserted `#include <com_err.h>`, removed that.

prolic:
Thanks, compiling works now. Will do a little more testing and let you know these days.

prolic:
Confirm mode seems not to work perfectly. 

See my test script here: https://gist.github.com/prolic/c631b4750e70f1218c46

Executing with: `while php ./confirm_test.php; do :; done`

Results in:

    string(3) "foo"
    string(3) "foo"
    string(3) "foo"
    string(3) "foo"
    string(3) "foo"
    string(3) "foo"
    PHP Fatal error:  Uncaught Error: Call to a member function getBody() on boolean in confirm_test.php:29
    Stack trace:
    #0 {main}
      thrown in confirm_test.php on line 29`

prolic:
Also when I cosume a message from the same channel in confirm mode, I get this exception:
Library error: unexpected protocol state

With phpamqplib I don't have this problem.

tests/amqpbasicproperties.phpt
pinepain: This is amqp properties that for some reason was mixed with envelope (see it [in java](http://www.rabbitmq.com/releases/rabbitmq-java-client/v3.3.0/rabbitmq-java-client-javadoc-3.3.0/com/rabbitmq/client/Envelope.html)) fields. `basic.return` method holds propertis and no enveloper, so to not confuse userland, `AMQPBasicProperties` introducd and now `AMQPMessage` extends it. Later we'll get rid of `AMQPMessage` or at least make it looks more logical.

pinepain:
> Confirm mode seems not to work perfectly.
> 
> See my test script here: https://gist.github.com/prolic/c631b4750e70f1218c46
> 
> Executing with: while php ./confirm_test.php; do :; done
> 
> Results in:
> 
> ```php
> string(3) "foo"
> string(3) "foo"
> string(3) "foo"
> string(3) "foo"
> string(3) "foo"
> string(3) "foo"
> PHP Fatal error:  Uncaught Error: Call to a member function getBody() on boolean in confirm_test.php:29
> Stack trace:
> #0 {main}
>   thrown in confirm_test.php on line 29`
> ```
> 

I don't get this code:

```php
$channel1->confirmSelect();
$exchange->publish('foo');
$msg = $queue->get();
var_dump($msg->getBody());
```

You entering confirm mode, but after publishing you don't wait for confirms with `AMQPChannel::waitForConfirm()` method, which doesn't makes much sense. You can look into relevant test file from this PR for example how to work with publish confirms.

Then you call `AMQPQueue::get()` method which is by-design may return `null` and call method on it result. While I follow your idea, it's not a quite safe way to operate.

pinepain:
> Also when I cosume a message from the same channel in confirm mode, I get this exception:
> Library error: unexpected protocol state
> 
> With phpamqplib I don't have this problem.
> 

Please, post reproducible example of this weird behavior and a way of how you do it in php-amqplib.

prolic:
My bad, I didn't used the waitForConfirm()-method. I'll try to give you more feedback soon.

pinepain:
That's fine, thank you for testing and feedback!

prolic:
I just got a segmentation fault when using your branch together with php 7 and xdebug. Don't have the time to get this isolated now, but I hope I can deliver a test script tomorrow. Do you want a gdb backtrace?

pinepain:
Can you post reproducible example of that segfault? Gdb trace may come in handy, but only in case I can't get it myself.

stubs/AMQPChannel.php
pinepain: Microseconds can be passed as a float part (after dot). Do we need to be microsec based? Overall precision is not good at all no, e.g. you can start waiting for 1.5 and it can be even 2.0 sec or smth, small values are not very precise, so I don't think we'll have any benefit from moving to microsec.

pinepain: But good note about guessing, I missed parameter docs.

prolic:
When creating a single test script it always wants to work.
So please try to clone from my repository and run phpunit with coverage-html on php 7.

see: https://gist.github.com/prolic/cd32a2469eda74bb005c

pinepain:
@prolic thanks for catching! Can you check with the submitted fix? My output now shows:

```php
PHPUnit 5.2.12 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.0.3 with Xdebug 2.4.1-dev
Configuration: /home/vagrant/php-amqp/HumusAmqp/phpunit.xml.dist

.........................................S....................... 65 / 76 ( 85%)
...........                                                       76 / 76 (100%)

Time: 41.57 seconds, Memory: 14.00Mb

There was 1 skipped test:

1) HumusTest\Amqp\PhpAmqpLib\ChannelRecoverTest::it_recovers
channel recover test not yet working for php amqp lib

/home/vagrant/php-amqp/HumusAmqp/tests/PhpAmqpLib/ChannelRecoverTest.php:42

OK, but incomplete, skipped, or risky tests!
Tests: 76, Assertions: 188, Skipped: 1.

Generating code coverage report in HTML format ... done
```

prolic:
Thanks, works!
I am about to move right now, so I cannot provide much more feedback the next 2 weeks.

pinepain:
ping @lstrojny what our roadmap on this PR? What about rolling new stable from current master and then merge it and publish new beta?

lstrojny:
@pinepain just back from vacation. Suggestion makes a lot of sense.

c-datculescu:
Hello. As mentioned in the publisher acknowledgements issue, i have ran into some issues involving fast publishing (16k/sec), where i get an exception with message "Unexpected server method". Didn't have time to drill more into it, but i will come back with more details as i have them.

lstrojny:
Merged into master, targetting 1.8.0

PR #207, opened 2016-03-05
return true on all set*-methods in AMQPExchange

prolic:


pinepain:
It was already discussed earlier: unless setter method may fail, it considered void. Setter methods that do remote calls, e.g. `AMQPChannel::setPrefetchCount()` which do `basic.qos` call under the hood may fail, and when it fail it throws exception. 

If there any violations of this rule in current codebase, it still so for BC saving reason and is subject to change in next major release, if any.

pinepain:
What is the reason to do this?

prolic:
I just noticed, that other methods return true after settings, f.e. here: https://github.com/pdezwart/php-amqp/blob/master/amqp_exchange.c#L280

Addtionally the return value is documented in the stubs, see here: https://github.com/pdezwart/php-amqp/blob/master/stubs/AMQPExchange.php#L225

So either we have to add the return value on the setters or update documentation.

lstrojny:
Thanks for pointing out the inconsistency, that’s indeed incorrect. All setter should be void methods.

prolic:
Ok, I'll open a new PR

prolic:
Btw: what about this one?

https://github.com/pdezwart/php-amqp/blob/master/amqp_connection.c#L961

prolic:
new PR here: https://github.com/pdezwart/php-amqp/pull/208

pinepain:
@prolic `php_amqp_disconnect_force()` throws exception in case it fail, so the only result that `AMQPConnection::setReadTimeout()` may return is `true`, if it fail it throws `AMQPConnectionException` exception.

PR #208, opened 2016-03-06
all setters return void

prolic:


pinepain:
Hi. Thanks for bringing attentions to API inconsistency. I appreciate your efforts to make this projects better.

This PR changes public API, so it breaks BC, which MAY break user code or tests and so on. While it is highly unlikely, it does break BC and by semantic versioning it should be a major release. We do aware about API clearance, but the last thing I want is to silently broke undefined number of applications in minor or even patch release. Even if we keep in mind that user should not rely on setters result. They do. I saw it in realworld application I had to maintain and fix and I see it from issues here and in related projects. We have large legacy codebase and no one happy from that API is weird at some placeses. I'm the first one of them. But breaking BC just to have cleaner API is not what I can understand.

I see your point, but while inconsistency is a bad thing, breaking BC is not much better.

pinepain:
@prolic Is there any practical reason to void/unvoid setters? Do you faced with some weird behavior caused by void/non-void setter?

prolic:
Ok, I understand. No there's no particular problem. I'll just submit another PR updating the stubs only.

pinepain:
Updating stubs with what? Specifying that non-void methods are void? That might be a good point to start, but then we'll probably should add `@throws` tag too when it matters.

PR #210, opened 2016-03-06
update stubs

prolic:


pinepain:
This small change is a really good start towards making our API more cleaner. Thanks!

Just an advice which I got long-long time ago: commit message starts with capital letter, be specific and short. I don't want to sound like I'm teaching you or something, just sharing good practices:

 - [Writing good commit messages](https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
 - [How to Write a Git Commit Message](http://chris.beams.io/posts/git-commit/)
 
there are another recommendations, but these two are my favorite. 

As to tests, when you edit file that doesn't affect test scope you may want to add `[skip ci]` to commit message to do not trigger build. In our case we have huge build matrix so we definitely do not want to flood travis-ci.

P.S.:
*(from my previous message, just to not to lose the idea)*
> This might be a good point to start, but then we'll probably should add relevant `@throws` tag too when it matters.

stubs/AMQPExchange.php
```diff
@@ -211,7 +211,7 @@ public function setArguments(array $arguments)
      *                       AMQP_DURABLE, AMQP_PASSIVE
      *                       (and AMQP_DURABLE, if librabbitmq version >= 0.5.3)
      *
-     * @return boolean True on success or false on failure.
+     * @return void
```
pinepain: This is a hollywar topic, but in the face of [Void Return Type](https://wiki.php.net/rfc/void_return_type), shouldn't we just drop any `@return` tag at all if method/function just doesn't (or shouldn't) explicitly return any result? I see that `void` clearly say that one should not relay on function result while not `@return` tag is not clear at this point.

prolic:
I am open for your suggestions on how to improve things. I just noticed the inconsistency and wanted to help :)

pinepain:
Yeah, it needs some plumbing here and there. Thanks.

lstrojny:
Thanks!

PR #218, opened 2016-04-26
add support for pkg-config

remicollet:
Since fe6920b4d107a2f73514f0a99d8233d20931d240 configure fails

    checking for amqp files in default path... found in /usr
    checking for librabbitmq version... NONE
    configure: error: could not determine librabbitmq version

In config.log

    configure:4058: cc -o conftest -I/usr/include  -Wl,-z,relro -specs=/usr/lib/rpm/redhat/redhat-hardened-ld conftest.c  >&5
    /usr/bin/ld: /tmp/ccggcfRK.o: relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
    /tmp/ccggcfRK.o: error adding symbols: Bad value
    collect2: error: ld returned 1 exit status

Adding "-fPIC" to CFLAGS indeed work.

This PR is about using librabbitmq.pc (provided in standard upstream installation) which seems much more simple (and shoud manage all cases, as pkg-config output is usally reliable, else it is an upstream bug)

P.S. only the 1st part of the commit is relevant, everything else is only "indent"

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -24,85 +24,111 @@ if test "$PHP_AMQP" != "no"; then
 	dnl # --with-amqp -> check with-path
 
 	SEARCH_FOR="amqp_framing.h"
+	AC_PATH_PROG(PKG_CONFIG, pkg-config, no)
 
-	AC_MSG_CHECKING([for amqp files in default path])
-	if test "$PHP_LIBRABBITMQ_DIR" != "no" && test "$PHP_LIBRABBITMQ_DIR" != "yes"; then
-		for i in $PHP_LIBRABBITMQ_DIR; do
-			if test -r $i/include/$SEARCH_FOR;
-				then
-				AMQP_DIR=$i
-				AC_MSG_RESULT(found in $i)
-				break
-			fi
-		done
+	if test "$PHP_LIBRABBITMQ_DIR" = "yes" -a -x $PKG_CONFIG; then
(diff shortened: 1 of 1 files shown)


remicollet:
Of course, nice to have in master and in the 1.7 branch

lstrojny:
Looks good 👍 

I wait for the build to pass until I merge into master and 1.7

pinepain:
It's probably me who broke build with fe6920b, but I can't get what exactly caused failure there. @remicollet can you suggest what did I broke so I can learn from it?

remicollet:
@pinepain the AC_TRY_RUN fails (see above output)

A possible fix is

     CFLAGS="$CFLAGS -I$AMQP_DIR/include"

pinepain:
Thanks! I even didn't suggest AC_TRY_RUN as a point of failure. Thanks again for pointing me on that and for the fix. Should this be added to config.m4 or fix you provides covers this case?

remicollet:
This PR use pkg-config when possible, but old code is still used as a fallback.
I will wait for travis result, to see if something missing (I just think it can fail if old version don't have .pc, but afaik, was already present in 0.5.2)

remicollet:
Travis was fine, but don't use pkg-config as library path is given as configure option.

I add a second commit for the CFLAGS

lstrojny:
Thank you @remicollet !

Issue #229, opened 2016-06-06
No message in queue after consume timeout and reject with requeue


prolic:
In words:

I send 2 messages to an exchange, they are routed to a queue.
I consume messages from there, without acknowledgement.
After specified read timeout the queue throws a timeout exception.
After consumer timeout I nack all and requeue them.

Then I need to call $queue->cancel($consumerTag), without this the messages are still delivered to the queue, even if it's no longer consuming (there was a consumer timeout exception !) - why is that? I don't know. But when this is the rule, I can live with it.

But when I now access the queue and try to fetch a single messages, the message is very often not there (you need to run the test script multiple times with different results).

Now to code:

    $c = new AMQPConnection(['read_timeout' => 1, 'write_timeout' => 1]);
    $c->connect();
    
    $ch = new AMQPChannel($c);
    
    $ex = new AMQPExchange($ch);
    $ex->setName('test-exchange');
    $ex->setType('direct');
    $ex->declareExchange();
    
    $q = new AMQPQueue($ch);
    $q->setName('test-queue');
    $q->setFlags(AMQP_DURABLE);
    $q->declareQueue();
    $q->bind('test-exchange');
    
    $q->purge(); // only for running this script multiple times
    
    $ex->publish('message #1');
    $ex->publish('message #2');
    $cnt = 0;
    
    $consumerTag = uniqid();
    $lastDeliveryTag = null;
    
    try {
        $q->consume(
            function (AMQPEnvelope $envelope) use (&$lastDeliveryTag) {
                var_dump($envelope->getBody());
                $lastDeliveryTag = $envelope->getDeliveryTag();
            },
            AMQP_NOPARAM,
            $consumerTag
        );
    } catch (\Exception $e) {
        var_dump($e->getMessage());
    }
    
    $q->nack($lastDeliveryTag, AMQP_MULTIPLE|AMQP_REQUEUE);
    $q->cancel($consumerTag); // why is this needed at all?
    
    sleep(2); // even with sleep I receive boolean instead of object of $q->get()
    
    $m = $q->get();
    var_dump(gettype($m));

No let's change the test-script a little:

    $c = new AMQPConnection(['read_timeout' => 1, 'write_timeout' => 1]);
    $c->connect();

    $ch = new AMQPChannel($c);

    $ex = new AMQPExchange($ch);
    $ex->setName('test-exchange');
    $ex->setType('direct');
    $ex->declareExchange();

    $q = new AMQPQueue($ch);
[...]

PR #250, opened 2016-10-23
Add AMQPChannel::close() method

pinepain:
This PR adds `AMQPChannel::close()` method. No BC changes or whatsoever.

This method was discussed many times in issues, the latest one could be found [here](https://github.com/pdezwart/php-amqp/issues/229).

### DIFF
--- a/amqp_channel.c
+++ b/amqp_channel.c
@@ -429,6 +429,22 @@ static PHP_METHOD(amqp_channel_class, isConnected)
 }
 /* }}} */
 
+/* {{{ proto bool AMQPChannel::close()
+Close amqp channel */
+static PHP_METHOD(amqp_channel_class, close)
+{
+    amqp_channel_resource *channel_resource;
+
+    PHP_AMQP_NOPARAMS();
+
+    channel_resource = PHP_AMQP_GET_CHANNEL_RESOURCE(getThis());
+
+    if(channel_resource && channel_resource->is_connected) {
+        php_amqp_close_channel(channel_resource TSRMLS_CC);
+    }
+}
+/* }}} */
+
 /* {{{ proto bool amqp::getChannelId()
 get amqp channel ID */
 static PHP_METHOD(amqp_channel_class, getChannelId)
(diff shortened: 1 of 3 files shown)


jippi:
This is exactly what we need too ! :)

pinepain:
@jippi glad to hear. Feedback is always welcomed here, so if this extension misses something or has a bug you could report and issue and someone will take care of it. Sure, PRs are welcomed even more 😄.

prolic:
Will test tomorrow.

amqp_channel.c
```diff
@@ -1117,6 +1136,7 @@ ZEND_END_ARG_INFO()
 zend_function_entry amqp_channel_class_functions[] = {
 		PHP_ME(amqp_channel_class, __construct, 	arginfo_amqp_channel_class__construct,		ZEND_ACC_PUBLIC)
 		PHP_ME(amqp_channel_class, isConnected, 	arginfo_amqp_channel_class_isConnected,		ZEND_ACC_PUBLIC)
+		PHP_ME(amqp_channel_class, close, 			arginfo_amqp_channel_class_close,			ZEND_ACC_PUBLIC)
```
jippi: nitpicking, indent is off for `arginfo_amqp_channel_class_close` :)

jippi:
@pinepain thanks! :)

Currently my 2 challenges is 

- be able to break out of the blocking consume() somehow (closing the connection causes segfault :))
- being able to access the number of prefetched messages on the channel (compared to the desired number, useful when gracefully shutting down and consumer tag has been canceled)

pinepain:
@jippi To leave blocking `consume()` you can set `read_timeout` connection option to some sane value which fits your needs and then catch timeout exception. As to segfault, can you please report issue with reproducible example please?

At this moment php-amqp uses rabbitmq-c and unfortunately it's not possible to get to prefetched messages.

amqp_channel.c
pinepain: thanks!

jippi:
@pinepain the consume is solved exactly like that, though it feels like a hack ;) 

Will open separate issue later

prolic:
tested and works like a charm!

prolic:
thanks @pinepain and @lstrojny 

now release it as 1.8.0 ? :)

PR #251, opened 2016-10-24
Set server error code on thrown exception

pinepain:
This PR pass AMQP server error code to thrown exceptions (when applicable) so it is now aavailable as`Exception::getCode()`.

This PR tends to be BC safe. Exception codes was not in use before so it's safe to start using them now. However, users should first run tests before upgrading just in case they relied on zero exception codes.


jippi:
nice! good job

lstrojny:
Looks good!

prolic:
Thanks a lot!

1.8.0 release please? :-)

PR #252, opened 2016-10-24
Remove obsoleted and non-supported versions

pinepain:
In this PR we get rid of non-supported PHP 5.3 and 5.4. While we don't drop support of it in code, extension can potentially work on them.

Also librabbitmq 0.5.2 removed from PHP 7.0 build matrix as it way too old and in distros and  package repositories (remirepo on rhel and ondrej on ubuntu), it makes no sense to test against it.

At this time we run 28 (!) build, which is a bit too high. With this changes we reduce number to 18.


lstrojny:
Looks good!

PR #254, opened 2016-10-25
Add PHP 7.1 to build martrix

pinepain:
Reformat travis config and add PHP 7.1 to build and allow it to fail.

Also provide fix to have green build.


lstrojny:
Thank you!

PR #266, opened 2017-01-18
Add myself to developers list as I finally got pecl accont [skip ci]

pinepain:
https://pecl.php.net/user/pinepain


lstrojny:
Yay!

PR #270, opened 2017-02-27
Remov PHP 5.5 from travis build matrix

pinepain:
As PHP 5.5 EOL for a long time (http://php.net/supported-versions.php) and PHP 5.6 provides security fixes only, let's remove PHP 5.5 from travis build matrix at this point to abuse travis infrastructure less. I guess there was changes since we removed 5.3 from which prevents this extension get build in PHP 5.3+, so *it won't have any negative impact on those who haven't adopted more recent PHP versions* for various reasons.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ env:
 
 matrix:
   allow_failures:
-    - php: next
+    - php: nightly
 
   include:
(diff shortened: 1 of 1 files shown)


lstrojny:
Makes sense

PR #282, opened 2017-06-08
Add consumer tag

### LINKED ISSUE #281: Consume from multiple queues. There is no way to tell what message from what queue came.
Given the example [amqpqueue_consume_multiple](https://github.com/pdezwart/php-amqp/blob/master/tests/amqpqueue_consume_multiple.phpt) there is no way to tell what was the exact queue the message came from. The [envelop.consumer_tag](https://github.com/alanxz/rabbitmq-c/blob/master/librabbitmq/amqp.h#L2218) is not taken into account nor [set to message](https://github.com/pdezwart/php-amqp/blob/master/amqp_envelope.c#L61). The second queue argument is always the object the `consume` [method was called on](https://github.com/pdezwart/php-amqp/blob/master/amqp_queue.c#L633). 

Solutions:

1. Add a consumer_tag to the message so we can do assumptions based on it inside the callback.
2. Maintain internally a map consumer_tag -> queue object. And use the correct queue object as the second argument.

The first one is easier to do (If it is ok I can send a PR for it) but the second one is more cleanish.

pinepain:
This PR adds consumer to envelope and introduce consumer tag to queue map to pass proper queue to consumer callback which should solve long-time problem when we pass to consumer a queue it was invoked on.

Changes to what queue is passed to consumer callback **potentially introduce BC break in rare cases**: before we always passed a queue consumer callback was invoked from (which is not quite right when multiple consumers on different queues running), but now proper queue will be passed. As previous behavior was error-prone in many ways and was not recommended to rely on (see discussions here in issue), I guess fixing that is least of two evils.

Closes #281


lstrojny:
Looking good! Thank you

makasim:
Could this be tagged? I've been using it for some time, so far so good.

lstrojny:
@makasim indeed, will roll a release this week.

PR #284, opened 2017-06-10
Fix handling channel consumers

pinepain:
This is a set of fixes for #282. Most notable change is that when envelope queue not found, new `AMQPEnvelopeException` thrown. Actual envelope stored in `AMQPEnvelopeException::$envelope` property. This will allow to resume orphaned message handling.

Behavior we introduced in #282 and improve in this PR to bail out when no consumer queue exists **is purely a BC**, so it should be definitely reflected in release changelog. Probably, a release with this changes should be at leas minor, it doesn't look for me like a patch. As to bumping major release version, i guess we are not there yet, however, by semantic versioning we are clearly there (actually, by semver we wold end with a major version greater then firefox version).


pinepain:
@makasim can you please look into this and perhaps test a bit?

makasim:
@pinepain It has been working fine for two days on my local. No crashes or other odd behaviors.

pinepain:
Thanks for testing!

pinepain:
@lstrojny Do we have any roadmap for this fixes? I'm worrying that people may start using existent master branch which is leaky.

lstrojny:
LGTM

makasim:
@pinepain @lstrojny any ETA on release?

lstrojny:
@makasim release is finally done!

makasim:
@lstrojny thank you very much

PR #291, opened 2017-08-04
Add a link to amqp interop compatible transport.

makasim:


pinepain:
Nice work, thanks!

I'd suggest to put abstractions and libraries into separate section (there would be more than one, I hope). Also, you may add `[skip ci]` to commit message as we don't run tests on readme file 😄 yet.

lstrojny:
Agreed, separate section makes sense. Otherwise I like the idea of cross-pollination in the PHP AMQP space 👍

makasim:
btw, there are RabbitMQ official tutorials adopted to amqp interop. They are working for amqp ext as well 

https://github.com/php-enqueue/enqueue-dev/tree/master/pkg/amqp-lib/tutorial

makasim:
@pinepain @lstrojny Should I do something more to get it merged?

lstrojny:
@makasim sorry for the late response. I would like to add a "Related libraries" section instead and as part of that section mention the interop project

makasim:
@lstrojny done

lstrojny:
Thanks for your contribution and your patience!

PR #295, opened 2017-11-10
Update vagrant & travis provision.

odolbeau:
Support of autoreconf have been removed from rabbitmq-c. Use cmake instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,8 +54,8 @@ matrix:
       env: LIBRABBITMQ_VERSION=v0.5.2 TEST_PHP_ARGS=-m
 
 before_install:
-  - sh provision/install_rabbitmq-c.sh ${LIBRABBITMQ_VERSION}
-  - if [ "${COVERITY_SCAN_BRANCH}" != 1 ]; then phpize && ./configure  --with-librabbitmq-dir=$HOME/rabbitmq-c && make ; fi
+  - sudo ./provision/install_rabbitmq-c.sh ${LIBRABBITMQ_VERSION}
+  - if [ "${COVERITY_SCAN_BRANCH}" != 1 ]; then phpize && ./configure && make ; fi
 
 install:
 
--- a/provision/install_rabbitmq-c.sh
+++ b/provision/install_rabbitmq-c.sh
@@ -2,23 +2,27 @@
 
 set -e
 
-echo Installing rabbitmq-c ...
-
+if [ "$#" -ne 1 ]; then
+  echo "Usage: $0 LIBRABBITMQ_VERSION" >&2
+  exit 1
+fi
 
 LIBRABBITMQ_VERSION=$1
 
-cd $HOME
+echo "Installing rabbitmq-c ..."
 
 if [ ! -d "$HOME/rabbitmq-c" ]; then
+  cd $HOME
   git clone git://github.com/alanxz/rabbitmq-c.git
+  cd $HOME/rabbitmq-c
 else
   echo 'Using cached directory.';
   cd $HOME/rabbitmq-c
   git fetch
 fi
 
-cd $HOME/rabbitmq-c
 git checkout ${LIBRABBITMQ_VERSION}
 
-git submodule init && git submodule update
-autoreconf -i && ./configure --prefix=$HOME/rabbitmq-c && make && make install
+mkdir build && cd build
+cmake ..
+cmake --build . --target install
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -69,7 +69,7 @@ sudo service nginx restart
 sudo cp -f /usr/share/nginx/html/index.html /var/www/html/index-nginx.html
 
 # Install and configure RabbitMQ
-wget -qO - http://www.rabbitmq.com/rabbitmq-signing-key-public.asc | sudo apt-key add -
+wget -qO - https://www.rabbitmq.com/rabbitmq-release-signing-key.asc | sudo apt-key add -
 sudo add-apt-repository 'deb http://www.rabbitmq.com/debian/ testing main'
 sudo apt-get update
 #sudo apt-get install --only-upgrade -y rabbitmq-server
@@ -81,7 +81,11 @@ sudo service rabbitmq-server restart
 # Note: it may be good idea to checkout latest stable rabbitmq-c version, but master branch for dev reasons also good
 cd ~
 git clone -q git://github.com/alanxz/rabbitmq-c.git
-cd rabbitmq-c && autoreconf -i && ./configure && make && sudo make install
+cd rabbitmq-c
+sudo apt-get install -y cmake
+mkdir build && cd build
+cmake ..
+sudo cmake --build . --target install
 # or install packaged version:
 #sudo apt-get install -y librabbitmq1 librabbitmq-dev librabbitmq-dbg
 


odolbeau:
Works great for the vagrant vm.

Same provision script for travis works great on swarrot (see: https://github.com/swarrot/swarrot/blob/master/tests/bin/install_rabbitmq-c.sh)

Looks like it's not working here due to the `sudo: false` in the `.travis.yml` I think.
Do you see any way to avoid this problem?

provision/install_rabbitmq-c.sh
```diff
@@ -1,24 +1,33 @@
...
 
 set -e
+set -x
 
-echo Installing rabbitmq-c ...
+# More information about this script here
+# * https://github.com/alanxz/rabbitmq-c/blob/master/.travis.yml
 
+echo Installing clang ...
+wget -O - http://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
```
lstrojny: Let’s use https here

provision/install_rabbitmq-c.sh
```diff
@@ -1,24 +1,33 @@
...
+apt-add-repository "deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-3.9 main"
```
lstrojny: … and here

provision/install_rabbitmq-c.sh
```diff
@@ -1,24 +1,33 @@
...
+apt-get -q update;
```
lstrojny: Superfluous semicolon

provision/install_rabbitmq-c.sh
```diff
@@ -1,24 +1,33 @@
...
+apt-get install -y clang-3.9 libpopt-dev;
```
lstrojny: Superfluous semicolon

provision/install_rabbitmq-c.sh
```diff
@@ -1,24 +1,33 @@
...
 
-LIBRABBITMQ_VERSION=$1
+echo Installing rabbitmq-c ...
```
lstrojny: Please quote the message

.travis.yml
```diff
@@ -54,7 +54,7 @@ matrix:
       env: LIBRABBITMQ_VERSION=v0.5.2 TEST_PHP_ARGS=-m
 
 before_install:
-  - sh provision/install_rabbitmq-c.sh ${LIBRABBITMQ_VERSION}
+  - sh -c 'provision/install_rabbitmq-c.sh'
```
pinepain: Is it intentional that `LIBRABBITMQ_VERSION` removed from args?

provision/install_rabbitmq-c.sh
pinepain: Could you clarify a bit what `llvm` is used here for?

provision/install_rabbitmq-c.sh
pinepain: And what is `clang` supposed to be used?

.travis.yml
odolbeau: The argument wasn't used in the script which directly use the environment variable.

odolbeau:
I copy / pasted the `before_install` part from rabbitmq-c ([here](https://github.com/alanxz/rabbitmq-c/blob/master/.travis.yml#L51-L56)).
Copy / pasting without thinking is bad! :/

`clang` could be used instead of `gcc` according to the README. `gcc` is fine, I removed this part.
`libopt` is a **optional** dependency => I removed it.

.travis.yml
odolbeau: My bad, it was using this argument. I can add it back if you prefer.

.travis.yml
```diff
@@ -54,7 +54,7 @@ matrix:
       env: LIBRABBITMQ_VERSION=v0.5.2 TEST_PHP_ARGS=-m
 
 before_install:
-  - sh provision/install_rabbitmq-c.sh ${LIBRABBITMQ_VERSION}
+  - sh -c './provision/install_rabbitmq-c.sh'
```
pinepain: I still think that `${LIBRABBITMQ_VERSION}` was removed by accident. Or did I miss something? 

The initial idea was that we pass desired version to install script and later read it from `$1`.

.travis.yml
```diff
@@ -54,7 +54,7 @@ matrix:
       env: LIBRABBITMQ_VERSION=v0.5.2 TEST_PHP_ARGS=-m
 
 before_install:
-  - sh provision/install_rabbitmq-c.sh ${LIBRABBITMQ_VERSION}
+  - sudo ./provision/install_rabbitmq-c.sh
```
pinepain: Ughh, commenting on this for the third time, I still can't get used to new GH review features or my comments get lost somehow. 

It looks like `LIBRABBITMQ_VERSION` was removed by accident as it value later retrieved in install script as `$1`.

Also, I'm not entirely sure that sudo is required to run this script. The install script use `sudo` within for commands that need it.

pinepain:
@odolbeau that's ok, build setting for CI are not the most obvious part.

Since you are there, could I ask you to alter provision script too - https://github.com/pdezwart/php-amqp/blob/b1e12ee93c3f9693c177ee58444582205d63d581/provision/provision.sh#L81-L84

.travis.yml
odolbeau: I removed the argument to use directly the `LIBRABBITMQ_VERSION` environment variable inside the script.
I can add the argument back if you prefer.

Without the sudo, here is the error when running `cmake`: https://travis-ci.org/swarrot/swarrot/jobs/300596815#L706 (it's on swarrot but it's the same install script)

odolbeau:
@pinepain I already updated the provision.sh script

.travis.yml
```diff
@@ -54,8 +54,8 @@ matrix:
       env: LIBRABBITMQ_VERSION=v0.5.2 TEST_PHP_ARGS=-m
 
 before_install:
-  - sh provision/install_rabbitmq-c.sh ${LIBRABBITMQ_VERSION}
-  - if [ "${COVERITY_SCAN_BRANCH}" != 1 ]; then phpize && ./configure  --with-librabbitmq-dir=$HOME/rabbitmq-c && make ; fi
+  - sudo ./provision/install_rabbitmq-c.sh
+  - if [ "${COVERITY_SCAN_BRANCH}" != 1 ]; then phpize && ./configure && make ; fi
```
odolbeau: I removed the `--with-librabbitmq-dir` option cause the rabbitmq-c library is now installed globally.

odolbeau:
OK, travis is still running but it looks like tests are green!
Sorry for all updates I did during your review.
This PR looks good to me right now, feel free to comment it. :)

.travis.yml
pinepain: I guess `${LIBRABBITMQ_VERSION}` was removed by accident. Or there was any other changes to build system I missed?

odolbeau: `${LIBRABBITMQ_VERSION}` is used to access to an environment variable.

Before: this environment variable was passed as an argument in the installation script
Now: The script directly use the environment variable

As said, if you prefer I can update the script to use an argument instead of the environment variable directly.

pinepain: The idea to install rabbitmq-c locally is to avoid using `sudo` in CI so that we could utilize container-based travis build env, [which is faster](https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/). I guess we already have `sudo: false` in `.travis-ci.yml` so unless there is hard requirement to install globally, I would avoid sudo and go with local install.

pinepain: @odolbeau oh, I'm sorry, missed that. I'd suggest to pass version explicitly. The rationale for that is that this script is copied time to time by others for their CI and for internal  purposes, so I'd prefer to make their life easier.

If you decide to bring argument passing back, could you also add this check to install script

```
if [ "$#" -ne 1 ]; then
  echo "Usage: $0 LIBRABBITMQ_VERSION" >&2
  exit 1
fi
```

odolbeau:
I added the argument back for the installation script.

Regarding the usage of `sudo`, it looks like the build time is similar to previous build ( Total time 3 hrs 7 min 34 sec ) which made me think it's still used container-based travis build env.

If you want to get rid of `sudo` & global install anyway, I'll be glad to have your help regarding the usage of `clang` to achieve this.

pinepain:
@odolbeau thanks a lot, looks decent! Let's merge!

provision/provision.sh
```diff
@@ -69,7 +69,7 @@ sudo service nginx restart
 sudo cp -f /usr/share/nginx/html/index.html /var/www/html/index-nginx.html
 
 # Install and configure RabbitMQ
-wget -qO - http://www.rabbitmq.com/rabbitmq-signing-key-public.asc | sudo apt-key add -
+wget -qO - http://www.rabbitmq.com/rabbitmq-release-signing-key.asc | sudo apt-key add -
```
lstrojny: Can we switch to https here as well?

odolbeau: Done.

lstrojny:
Thank you!

Issue #288, opened 2017-07-03
Extension license


pinepain:
php-amqp ext still use PHP license, though extension itself is neither PHP core extension nor relates to PHP/Zend organization, so it might be a good move to start using more popular and well-known license which explicitly covers all usage aspects.

MIT looks like one that just do the job well.

What php-amqp community think about this idea?

@lstrojny do you have any particular thoughts on topic?

### Contributors who need to agree to a license change

- [x] 23006 pinepain@gmail.com
- [ ] 2560 pdezwart@rubiconproject.com
- [x] 2398 lars@strojny.net
- [x] 1219 bkw@codingforce.com
- [ ]  255 sladkov@gmail.com
- [ ]  107 ab@php.net
- [x]   95 sqmk@php.net
- [x]   90 remi@famillecollet.com
- [ ]   88 dizeee@dizeee.ru
- [x]   87 lars.strojny@internations.org
- [ ]   46 vkartaviy@gmail.com
- [ ]   25 peter.hansen@rwth-aachen.de
- [x]   21 lstrojny@lstrojnypro.local
- [ ]   18 github@a.bbnt.me
- [ ]   14 artemjo@gmail.com
- [x]   10 fedora@famillecollet.com
- [ ]    6 saschaprolic@googlemail.com
- [ ]    5 kangur91@gmail.com
- [ ]    4 siad.ardroumli@googlemail.com
- [x]    4 kotlyar.maksim@gmail.com
- [ ]    1 sergsk@gmail.com
- [ ]    1 sergeyklay@users.noreply.github.com
- [ ]    1 lyrixx@lyrixx.info
- [ ]    1 kazuki_yasufuku@dwango.co.jp

lstrojny:
In hindsight I believe using MIT or the Apache license would have been better. Organising a license change is nothing I am personally interested in doing, I would agree to a license change though immediately.

lstrojny:
Here is how to find all contributors and how many lines of code are still 

```
23006 pinepain@gmail.com
2560 pdezwart@rubiconproject.com
2398 lars@strojny.net
1219 bkw@codingforce.com
 255 sladkov@gmail.com
 107 ab@php.net
  95 sqmk@php.net
  90 remi@famillecollet.com
  88 dizeee@dizeee.ru
  87 lars.strojny@internations.org
  46 vkartaviy@gmail.com
  25 peter.hansen@rwth-aachen.de
  21 lstrojny@lstrojnypro.local
[...]

PR #299, opened 2017-11-19
Use MIT license

pinepain:
As per #288


lstrojny:
@pinepain, unfortunately, I can’t just merge that since we would have to ask every contributor that still has code in the codebase for consensus.

pinepain:
It's a good idea to ask community, maybe doing that via the GitHub issue (#288) was not the best way how to handle that.

I did not find whete PHP License (http://php.net/license/3_01.txt) specifies whether all derivative versions must utilize the same license or where it specifies anything about license change terms.

What should be our further plans on this issue, if any?

Issue #300, opened 2017-11-19
Add Dockerfile for dev needs and for CI


pinepain:
It came up that segfault in #293 happened only in docker environment, so it make sense to add docker support to CI. 

Based on how popular docker is it's vital to offer Dockerfile for contributors and potential users who'd like to evaluate php-amqp. It may worth to offer docker compose with rabbitmq and maybe nginx and apache, however web servers are optional, imo.

lstrojny:
Fixed in e9c89383da09749b65786bb49a94db590a973f79

PR #302, opened 2017-11-19
Fix segfault when consumer tag canceled on queue that has no consumers

### LINKED ISSUE #293: \AMQPQueue::cancel() causes segfault.
I failed to come up with a small piece of code that reproduces the error but it does segfault in my tests [here](https://github.com/queue-interop/queue-spec/blob/c252543f6c1b0e026aed0561991d076e0b81ee75/src/Amqp/BasicConsumeShouldRemoveConsumerTagOnUnsubscribeSpec.php#L30) and [here](https://github.com/queue-interop/queue-spec/blob/c252543f6c1b0e026aed0561991d076e0b81ee75/src/Amqp/BasicConsumeUntilUnsubscribedSpec.php#L30). If I comment out [this line](https://github.com/php-enqueue/enqueue-dev/blob/69634d72ec7934a60cc4bff92b9df81cfee6d9b1/pkg/amqp-ext/AmqpContext.php#L336) tests obviously fail but there is no segfault. 

That's all I have for now. I'll try to come back with steps to repo.

pinepain:
Closes #293.

As for test, it still green in vagrant env for some reasons, see #300 for details how we can avoid such issues in the future by adding docker step to ci.

### DIFF
--- a/amqp_queue.c
+++ b/amqp_queue.c
@@ -947,6 +947,7 @@ static PHP_METHOD(amqp_queue_class, cancel)
 
 	zval *channel_zv = PHP_AMQP_READ_THIS_PROP("channel");
 	zval *consumers = zend_read_property(amqp_channel_class_entry, channel_zv, ZEND_STRL("consumers"), 0 PHP5to7_READ_PROP_RV_PARAM_CC TSRMLS_CC);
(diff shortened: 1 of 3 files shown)


pinepain:
The second commit is about first two bullets in Two Hard Problems in distributed systems:

> There are only two hard problems in distributed systems:
> 2. Exactly-once delivery
> 1. Guaranteed order of messages
> 2. Exactly-once delivery

lstrojny:
Thank you!

PR #303, opened 2017-11-19
Drop librabbitmq < 0.7.1 support

### LINKED ISSUE #301: Dropping librabbigmq < 0.7.1 support
Currently Ubuntu 14.04 LTS [has 0.4.1](https://packages.ubuntu.com/trusty/librabbitmq-dev), which we don't support anyway and 16.04 LTS [has 0.7.1](https://packages.ubuntu.com/xenial/librabbitmq-dev).

From php-amqp perspective we highly recommend to have librabbitmq-c > 0.6.0.

I propose to drop everything less then 0.7.1 from support to reduce build matrix and encourage people to use more recent software (bug fixes, security issues, new features, etc.).

I'd suggest to do this drop in next minor release, which I guess could be `1.10.z`.

Are there any objections?

What do you think @lstrojny?

pinepain:
Closes #301

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,6 @@ matrix:
       env: LIBRABBITMQ_VERSION=v0.8.0 TEST_PHP_ARGS=-m
     - php: 7.1
       env: LIBRABBITMQ_VERSION=v0.7.1 TEST_PHP_ARGS=-m
-    - php: 7.1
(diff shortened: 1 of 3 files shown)


pinepain:
Red build is because of bogus test with inconsistent output. It is fixed in #302. To make this build green just restart red steps (not nice, but it works).

lstrojny:
Thank you!

PR #307, opened 2017-12-21
Add cmake addons

odolbeau:
`cmake` is not installed by default anymore.


pinepain:
Thanks, good point. Anyway, it's nice to ensure that dependencies are installed regards whether they are supposed to be installed by default or not.

lstrojny:
Thank you!

Issue #316, opened 2018-05-02
the extension can not be recognized under PHP 5.3.X since version 1.9.x


yjqg6666:
The  support for PHP 5.3 has already been dropped? If so, the README.md could be updated.

$php -m
PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib64/php/modules/amqp.so' - /usr/lib64/php/modules/amqp.so: undefined symbol: _php_math_number_format_ex in Unknown on line 0

https://github.com/pdezwart/php-amqp/blob/v1.9.0beta1/amqp_timestamp.c#L61-L71 

The _php_math_number_format_ex API was introduced in PHP 5.4 by commit https://github.com/php/php-src/commit/d2263d482d58180377a3c5e848352dbf0421cb68.

pinepain:
Hi, thanks for bringing this up. AFAIR, PHP versions below 5.6 (all [EOLed](http://php.net/supported-versions.php) versions) as not supported. It may work but it's not guaranteed.

Do you mind to PR min PHP version check? You can use https://github.com/php-v8/php-ref/blob/4129b661b0ff9e60b291d2592de64a7abafbca1b/config.m4#L14-L18 and https://github.com/php-v8/php-ref/blob/4129b661b0ff9e60b291d2592de64a7abafbca1b/php_ref.h#L29-L32 as a reference.

yjqg6666:
Thanks for the reference. I just made a PR.

PR #317, opened 2018-05-03
Add PHP support check for #316 

yjqg6666:
Only PHP version >= 5.6 could compile. See #316

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,10 @@
 
 Object-oriented PHP bindings for the AMQP C library (https://github.com/alanxz/rabbitmq-c)
 
-This extension works both with PHP 5 and PHP 7 (both zts and non-zts).
 
 ### Requirements:
 
+ - PHP >= 5.6 (PHP 7 included) with either ZTS or non-ZTS version.
  - [RabbitMQ C library](https://github.com/alanxz/rabbitmq-c), commonly known as librabbitmq
    (since php-amqp>=1.6.0  librabbitmq >= 0.5.2 required, >= 0.6.0 recommended,
    see [release note](http://pecl.php.net/package/amqp/1.6.0)).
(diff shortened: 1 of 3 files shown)


pinepain:
Awesome, thanks! Would you mind to add a note to readme that PHP 5.6 required, e.g. to alter `This extension works both with PHP 5 and PHP 7 (both zts and non-zts).` to sth like `This extension works both with PHP >= 5.6 (including PHP 7) with both zts and non-zts versions.` or any other variations you will find reasonable.

yjqg6666:
@pinepain The README updated. I put it into the requirement section.

lstrojny:
Thank you!

ryandesign:
The pecl page still says it supports PHP 5.3 or later.

lstrojny:
The change has not yet been released, that’s why it still says 5.3

ryandesign:
Yes it has. [amqp 1.9.4 was released to pecl](https://pecl.php.net/package/amqp/1.9.4). It requires PHP 5.6 or newer, but the pecl page says it requires 5.3 or newer, which is wrong.

lstrojny:
I don’t think so. Have a look at the package.xml in 1.9.4: https://github.com/pdezwart/php-amqp/blob/12bab1199e91f9193d759aa4297dc5d1ceac071e/package.xml

```
    <dependencies>
        <required>
            <php>
                <min>5.3.0</min>
            </php>
            <pearinstaller>
                <min>1.4.0</min>
            </pearinstaller>
        </required>
    </dependencies>

```

lstrojny:
@ryandesign well, that’s why this changed in master (https://github.com/pdezwart/php-amqp/blob/master/package.xml)

```xml
    <dependencies>
        <required>
            <php>
                <min>5.6.0</min>
            </php>
            <pearinstaller>
                <min>1.4.0</min>
            </pearinstaller>
        </required>
    </dependencies>
```

ryandesign:
Ok great, thanks. I don't suppose there's a way to fix the pecl web site to indicate that 1.9.4 in fact requires 5.6, without having to repackage the tarball?

PR #318, opened 2018-05-04
add cmake support for code navigation in Clion

yjqg6666:
Regarding the  C STANDARD adopted, which version should be?


yjqg6666:
Reasonable considerations. Thanks.

yjqg6666:
I created a [gist](https://gist.github.com/yjqg6666/19c8d593963997255d46a066868da73c), just for reference.

PR #320, opened 2018-05-30

bhoehl:
@pinepain @lstrojny 

		resource->connection_state,
		params->vhost,
		params->channel_max,
		params->frame_max,
		params->heartbeat,
		&custom_properties_table,
==>		AMQP_SASL_METHOD_PLAIN, <==
		params->login,
		params->password
	);


amqp.c
```diff
@@ -172,6 +173,8 @@ static PHP_MINIT_FUNCTION(amqp) /* {{{ */
 	REGISTER_LONG_CONSTANT("AMQP_OS_SOCKET_TIMEOUT_ERRNO",	AMQP_OS_SOCKET_TIMEOUT_ERRNO,	CONST_CS | CONST_PERSISTENT);
 	REGISTER_LONG_CONSTANT("PHP_AMQP_MAX_CHANNELS",			PHP_AMQP_MAX_CHANNELS,			CONST_CS | CONST_PERSISTENT);
 
+    REGISTER_LONG_CONSTANT("AMQP_SASL_METHOD_PLAIN",		AMQP_SASL_METHOD_PLAIN,		CONST_CS | CONST_PERSISTENT);
```
pinepain: formatting: tabs vs spaces, as this project uses tabs, for consistency purpose let's keep using tabs too.

BTW, I'm having this issue all the time too =).

amqp_connection.c
```diff
@@ -1238,6 +1248,56 @@ static PHP_METHOD(amqp_connection_class, setVerify)
...
+
+/* {{{ proto amqp::setSaslMethod(mixed method)
+set sasl method */
+static PHP_METHOD(amqp_connection_class, setSaslMethod)
+{
+	zval *zvalMethod;
+	int method;
+
+	/* Get the port from the method params */
+	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "z", &zvalMethod) == FAILURE) {
```
pinepain: Let's just enforce it to be decimal and all conversion will be done by PHP

pinepain:
@bhoehl Thanks for contribution, looks decent. After minor cleanup and more eyes I'm more then happy to have it in the core too!

bhoehl:
@pinepain 
I have addressed the issues with tabs and spaces
also I have looked for the unit tests, but for some reason my commit didn´t fix it
is there a way to get a bit more details on those failures ?

pinepain:
@bhoehl thanks! Give another comment regards regards long a look (I see that you take `setPort` as a reference, but it also should be refactored, but later).

Regards broken tests, see travis build output (the red cross near the commit will get you there or just go to, say, https://travis-ci.org/pdezwart/php-amqp/builds/385629994)

PR #321, opened 2018-05-31

bhoehl:
@pinepain @lstrojny

resource->connection_state,
params->vhost,
params->channel_max,
params->frame_max,
params->heartbeat,
&custom_properties_table,
==>	AMQP_SASL_METHOD_PLAIN, <==
params->login,
params->password
);


bhoehl:
@pinepain @lstrojny 
I have recreated that PR to purge those commits and addressed the issues with message naming and unit tests. Please have a look.

amqp_connection.c
```diff
@@ -510,6 +512,14 @@ static PHP_METHOD(amqp_connection_class, __construct)
 		}
 	}
 
+	zend_update_property_long(this_ce, getThis(), ZEND_STRL("sasl_method"), INI_INT("amqp.sasl_method") TSRMLS_CC);
+
+		if (ini_arr && PHP5to7_ZEND_HASH_FIND(HASH_OF(ini_arr), "sasl_method", sizeof("sasl_method"), zdata)) {
```
pinepain: minor formatting: Unexpected indent

stubs/AMQPConnection.php
```diff
@@ -434,4 +434,20 @@ public function getVerify()
     public function setVerify($verify)
     {
     }
+
+    /**
+     * set authentication method
+     *
+     * @param int $method AMQP_SASL_METHOD_PLAIN | AMQP_SASL_METHOD_EXTERNAL
+     */
+    public function setSasslMethod($method)
```
pinepain: Double `s` in setSa**ss**lMethod()

bhoehl:
@pinepain 
addressed

PR #322, opened 2018-06-01

bhoehl:
@pinepain @lstrojny
- squashed commits

resource->connection_state,
params->vhost,
params->channel_max,
params->frame_max,
params->heartbeat,
&custom_properties_table,
==>	AMQP_SASL_METHOD_PLAIN, <==
params->login,
params->password
);


pinepain:
@bhoehl LGTM in general. Thank you for all your efforts!

As @lstrojny mentioned before, how feasible do you find adding integration test to demonstrate SASL auth usage?

BTW, you may maintain all your changes in a single PR, it makes it a bit easier to follow and review, as well as it keeps all reviews and discussions at a single place.

bhoehl:
@pinepain 
it ought to be not a big deal, as there are already test cases for ssl
so, using those and just enabling external auth ought to be already sufficient, without supplying a password
I will look into this soon

bhoehl:
@pinepain @lstrojny 
I have added integration tests to these ssl authentications
for some reason, they do only run in vagrant virtualbox
the test environment in travis seems to need different pathes
please test them locally and have a look, please
thanks!

tests/amqpconnection_ssl_by_cert.phpt
```diff
@@ -0,0 +1,156 @@
...
+you can validate authentication by looking at /var/log/rabbitmq/rabbit@vagrant-sasl.log
+connection <0.1104.0> (127.0.0.1:48898 -> 127.0.0.1:5671): user 'sasl-client.example.org' authenticated and granted access to vhost '/'
+*/
+$credentials = array(
+    'host' => 'rabbitmq.example.org',
+    'port' => 5671,
+
+    'verify' => true,
+
+    'cacert' => "/home/vagrant/php-amqp/provision/test_certs/testca/cacert.pem",
```
pinepain: (a note for myself) let's not using absolute route here. I guess sth like `__DIR__ . '../provision/....'` should do the job pretty well. This probably could/should be handled during further refactoring steps.

tests/amqpconnection_ssl_by_cert.phpt
```diff
@@ -0,0 +1,156 @@
...
+  string(63) "/home/vagrant/php-amqp/provision/test_certs/sasl-client/key.pem"
+  ["cert":"AMQPConnection":private]=>
+  string(64) "/home/vagrant/php-amqp/provision/test_certs/sasl-client/cert.pem"
+  ["verify":"AMQPConnection":private]=>
+  bool(true)
+  ["sasl_method":"AMQPConnection":private]=>
+  int(1)
+}
+
+Fatal error: Uncaught AMQPConnectionException: Socket error: could not connect to host. in /home/vagrant/php-amqp/tests/amqpconnection_ssl_by_cert.php:68
```
pinepain: Is it expected here?

bhoehl: @pinepain 
yes, it is, because we are connecting to localhost with host verification turned on
( verify = true )

bhoehl:
@pinepain 
what is your plan with this code ?
are you expecting me to change futher stuff, if so please let me know
Thanks!

bhoehl:
@pinepain @lstrojny 
I have adjusted integration tests to use relative path and a better excption handling

amqp_connection.c
```diff
@@ -85,17 +85,17 @@ zend_object_handlers amqp_connection_object_handlers;
...
+	}
 
 	return ZEND_HASH_APPLY_KEEP;
 }
 
 static PHP5to7_param_str_len_type_t php_amqp_get_connection_hash(amqp_connection_params *params, char **hash) {
 	return spprintf(hash,
 					0,
-					"amqp_conn_res_h:%s_p:%d_v:%s_l:%s_p:%s_f:%d_c:%d_h:%d_cacert:%s_cert:%s_key:%s",
+					"amqp_conn_res_h:%s_p:%d_v:%s_l:%s_p:%s_f:%d_c:%d_h:%d_cacert:%s_cert:%s_key:%s:%d",
```
lstrojny: Should be `…sasl_method:%d` for consistency

amqp_connection.c
```diff
@@ -1238,6 +1248,39 @@ static PHP_METHOD(amqp_connection_class, setVerify)
...
+	long method;
+
+	/* Get the port from the method params */
+	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "l", &method) == FAILURE) {
+		return;
+	}
+
+	/* Check the method value */
+	if (method != AMQP_SASL_METHOD_PLAIN && method != AMQP_SASL_METHOD_EXTERNAL) {
+		zend_throw_exception(amqp_connection_exception_class_entry, "Invalid sasl method given. Method must be AMQP_SASL_METHOD_PLAIN or AMQP_SASL_METHOD_EXTERNAL.", 0 TSRMLS_CC);
```
lstrojny: Message should be `Invalid SASL method given …`

amqp_connection.c
bhoehl: @lstrojny 
addressed

amqp_connection.c
bhoehl: @lstrojny 
addressed

bhoehl:
@lstrojny 
squashed commits

lstrojny:
@bhoehl thank you very much!

pinepain:
@bhoehl Great job, thanks a lot!

pinepain:
@bhoehl Hi, @lstrojny is the release manager, let's ask him. No objections from my side if you use it in a wild already (however i'm not currently using php-amqp, so pls, count this fact in).

bhoehl:
@pinepain @lstrojny Thanks for the info, yes we will use this in production systems. So, there is a real need for having a new release.

PR #341, opened 2019-02-27
Fix explicit null-string for $routing_key in Queue::bind() and Exchange::publish()

xodmqjw:
`Exchange::publish()` and `Queue::bind()` according to their interfaces use `null` value as a default value for `$routing_key`. And all is OK if use just:

`$exchange->publish($msg);`

But, if we use smth like this:

`$exchange->publish($msg, null, $flag, $attrs); // explicit null value for a $routing_key`

we'll get:

> PHP Fatal error:  Uncaught TypeError: AMQPExchange::publish() expects parameter 2 to be string, null given in ...

And in a case when we extend Exchange or Queue their methods looks like:

```
class Exchange extends \AMQPExchange
{
    public function publish($msg, $routingKey = null, ...)
    {
        parent::publish($msg, $routingKey, ...);// bug, we'll get TypeError here if someone calls child-method only with $msg
    }
}
```

Env:
PHP 7.3.2
Zend Engine v.3.3.2
PHP-AMQP v.1.9.4-1

### DIFF
--- a/amqp_exchange.c
+++ b/amqp_exchange.c
@@ -439,7 +439,7 @@ static PHP_METHOD(amqp_exchange_class, publish)
 
 	amqp_basic_properties_t props;
 
-	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "s|sla/",
+	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "s|s!la/",
 							  &msg, &msg_len,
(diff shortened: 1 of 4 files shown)


lstrojny:
Thank you. Could you add a test case as well?

xodmqjw:
> Thank you. Could you add a test case as well?

Done.

lstrojny:
Thank you!

PR #346, opened 2019-04-30
Pass params by value in AMQPConnection::__construct()

xodmqjw:
Inside AMQPConnection::__construct() type of params (read, write and connect timeouts) passed within an array is implicitly changed (convert_to_double()). Separating an array will help to avoid this problem and will prevent possible accidental changes of values in the future.


lstrojny:
Thank you!

Issue #355, opened 2019-12-02
Compile failure on php 7.4


monteiro:
After installing php 7.4 and run:

```
pecl install amqp
```

I got some compilation errors.

```
running: make
/bin/sh /private/tmp/pear/temp/pear-build-hugoqKJyG7/amqp-1.9.4/libtool --mode=compile cc  -I. -I/private/tmp/pear/temp/amqp -DPHP_ATOM_INC -I/private/tmp/pear/temp/pear-build-hugoqKJyG7/amqp-1.9.4/include -I/private/tmp/pear/temp/pear-build-hugoqKJyG7/amqp-1.9.4/main -I/private/tmp/pear/temp/amqp -I/usr/local/Cellar/php/7.4.0/include/php -I/usr/local/Cellar/php/7.4.0/include/php/main -I/usr/local/Cellar/php/7.4.0/include/php/TSRM -I/usr/local/Cellar/php/7.4.0/include/php/Zend -I/usr/local/Cellar/php/7.4.0/include/php/ext -I/usr/local/Cellar/php/7.4.0/include/php/ext/date/lib -I/usr/local/Cellar/rabbitmq-c/0.9.0_1/include  -DHAVE_CONFIG_H  -g -O2   -c /private/tmp/pear/temp/amqp/amqp.c -o amqp.lo
mkdir .libs
 cc -I. -I/private/tmp/pear/temp/amqp -DPHP_ATOM_INC -I/private/tmp/pear/temp/pear-build-hugoqKJyG7/amqp-1.9.4/include -I/private/tmp/pear/temp/pear-build-hugoqKJyG7/amqp-1.9.4/main -I/private/tmp/pear/temp/amqp -I/usr/local/Cellar/php/7.4.0/include/php -I/usr/local/Cellar/php/7.4.0/include/php/main -I/usr/local/Cellar/php/7.4.0/include/php/TSRM -I/usr/local/Cellar/php/7.4.0/include/php/Zend -I/usr/local/Cellar/php/7.4.0/include/php/ext -I/usr/local/Cellar/php/7.4.0/include/php/ext/date/lib -I/usr/local/Cellar/rabbitmq-c/0.9.0_1/include -DHAVE_CONFIG_H -g -O2 -c /private/tmp/pear/temp/amqp/amqp.c  -fno-common -DPIC -o .libs/amqp.o
[...]

PR #360, opened 2020-01-02
Replace `ulong` with `zend_ulong`

ahilles107:
use `zend_ulong` instead `ulong` - solves php7.4 compilation error


ahilles107:
Fixes https://github.com/pdezwart/php-amqp/issues/355

Issue #354, opened 2019-12-02
/tests/bug_gh147 fails on php7.4


ludoge:
```
---- EXPECTED OUTPUT
start
consuming
%s fatal error: Object of class %s could not be converted to string in %s on line %d
---- ACTUAL OUTPUT
start
consuming
Fatal error: Uncaught Error: Object of class AMQPEnvelope could not be converted to string in /home/ludoge/vendor/php-amqp/stretch/src-7.4-zts/tests/bug_gh147.php:27
Stack trace:
#0 [internal function]: {closure}()
#1 /home/ludoge/vendor/php-amqp/stretch/src-7.4-zts/tests/bug_gh147.php(30): AMQPQueue->consume()
#2 {main}
  thrown in /home/ludoge/vendor/php-amqp/stretch/src-7.4-zts/tests/bug_gh147.php on line 27
---- FAILED
```
This is due to https://wiki.php.net/rfc/tostring_exceptions

lstrojny:
Fixed now in master

PR #358, opened 2019-12-10
Skip failing tests for 7.4

ludoge:
See #354 and #355

PR #365, opened 2020-02-25
Fixed build on Travis CI

supersmile2009:
1. Fixed RabbitMQ installation. Since Xenial it's not available as a service with Travis and has to be installed with `apt`.
2. Fixed test `bug_gh147.phpt` on PHP 7.4. As of PHP 7.4 converting object to string throws an exception if it cannot be converted. Since error in PHP 7.3 and exception in PHP 7.4 is formatted differently, I catch both the error (with error handler function) and exception and only print the message.
This PR fixes issue https://github.com/pdezwart/php-amqp/issues/354 with a better solution than https://github.com/pdezwart/php-amqp/pull/358 (which simply skips the problematic test)
3. Disabled memleak check on PHP 7.4 as it reports false positives same as in PHP 7.3. I'll provide a PR with PoC of memleak tests re-enabled and explanation later.

We should finally have a green build here with these fixes :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,5 @@
 language: php
 
-sudo: false
-
-services:
-  - rabbitmq
-
 env:
   global:
     - NO_INTERACTION=1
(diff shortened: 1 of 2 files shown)


lstrojny:
Awesome, thank you!

Issue #366, opened 2020-03-03
Add support for global prefetch count to consumers


tezhm:
As per specification, the prefetch count for all consumers on a channel can be set via a flag (see https://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.qos.global) but in the amqp_channel.c it is hard coded to 0.

Would like support for this added so we can customise total number of messages consumed on a channel rather than per consumer.

PR #367, opened 2020-03-04
Adding global prefetch support (#366)

tezhm:
Adding support for the global flag (https://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.qos.global). As a note, consumer prefetch settings can be set in conjunction with the channel wide prefetch settings.

Through testing I noticed that setting consumer prefetch settings actually clears the channel wide settings (testing against RabbitMQ 3.8.2). I'm not sure if this being cleared from the client library or from the server. To workaround this I re-apply the channel wide settings after applying the consumer prefetch settings.

I didn't look into previous usage of the global flag so am unaware of backwards compatibility issues, mainly from new code which re-applies the global prefetch settings within the existing `setPrefetchCount` and `setPrefetchSize` methods.

From design point of view, wasn't sure if adding new methods or adding an optional bool flag to the `setPrefetchCount` and `setPrefetchSize` methods would be preferred. Both are backwards compat. I decided to add separate methods as global/consumer are different sets of properties. However, I added an optional bool flag to the `qos` method because didn't really like the `globalQos` method name. This can be changed.

TODO: unit tests


lstrojny:
Looks great! Could you add a few tests, if only for the setters?

tezhm:
Added some setter tests for the new methods. The prefetch size is not supported server side with RabbitMQ 3.8.2 so that test has a condition to skip if 504 -> NOT_IMPLEMENTED is returned.

Did you want me to add some more tests for setPrefetchCount and setPrefetchSize? There are tests implicitly hitting the code but nothing explicit. Could use some ini tests and setter getter stuff.

lstrojny:
> Did you want me to add some more tests for setPrefetchCount and setPrefetchSize? There are tests implicitly hitting the code but nothing explicit. Could use some ini tests and setter getter stuff. Also, could do some more thorough backwards compatibility checks with this stuff.

I didn’t initially intend that but now that you are mentioning it, it would be great.

tests/amqpchannel_set_global_prefetch_size.phpt
```diff
@@ -0,0 +1,37 @@
...
+            print "skip";
+        }
+    }
+}
+?>
+--FILE--
+<?php
+$cnn = new AMQPConnection();
+$cnn->connect();
+$ch = new AMQPChannel($cnn);
```
lstrojny: `var_dump`ing the default before setting one is probably healthy

tests/amqpchannel_set_global_prefetch_count.phpt
```diff
@@ -0,0 +1,26 @@
...
+if (!extension_loaded("amqp")) {
+    print "skip";
+}
+?>
+--FILE--
+<?php
+$cnn = new AMQPConnection();
+$cnn->connect();
+$ch = new AMQPChannel($cnn);
+$ch->setGlobalPrefetchCount(123);
```
lstrojny: `var_dump`ing the default before setting one is probably healthy

tezhm:
Added some more tests. Since this is a wrapper for the client side library, I'm not sure if we need to test behaviour of the server in the sense that amqp versions interpret the global flag as channel/connection, whereas rabbitmq interprets as consumer/channel (see https://www.rabbitmq.com/consumer-prefetch.html). I was thinking of testing backwards compatibility by checking that consumer settings are applied in those terms depending on server version.

Any input on this?

lstrojny:
> Any input on this?

Indeed, the current level of coverage is fine.

Thank you very much, great work. Will merge now!

PR #368, opened 2020-04-04
Avoid variable lengths arrays

cmb69:
Despite conforming to C99, variable length arrays are not supported by
MSVC, and maybe some other relevant compilers.  Therefore we hard-code
the maximum required space.

---

This would solve the Windows PECL [build failures](https://windows.php.net/downloads/pecl/releases/amqp/1.10.0/). An alternative would be to use do_alloca() and friends (similar to https://github.com/php/php-src/commit/9d31a42a30e944688c29aefc4bd0396ce395efe1), but this appears to be overkill in this case.

### DIFF
--- a/amqp_connection_resource.c
+++ b/amqp_connection_resource.c
@@ -534,7 +534,7 @@ amqp_connection_resource *connection_resource_constructor(amqp_connection_params
 	}
 
 	custom_properties_table.entries     = custom_properties_entries;
-	custom_properties_table.num_entries = sizeof(custom_properties_entries) / sizeof(amqp_table_entry_t);
+	custom_properties_table.num_entries = params->connection_name ? 2 : 1;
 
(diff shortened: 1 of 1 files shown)


lstrojny:
@cmb69 can you check the travis output, it seems to segfault. Happy to roll a dot-one release very quickly to get the Windows build fixed.

For a longer term solution, I have opened https://github.com/pdezwart/php-amqp/issues/369

cmb69:
Thanks for the ping! That segfault should be fixed now. I'll try to have a look at AppVeyor CI later today or tomorrow.

lstrojny:
@cmb69 thanks a lot!

And man it would be terrific to have a good Windows CI setup. Thank you very much already!

lstrojny:
v1.10.1 is released now with the fix

PR #370, opened 2020-04-05
Add AppVeyor CI support (closes #369)

### LINKED ISSUE #369: CI for Windows
I don't have a Windows build environment but I would like to ensure Windows compatibility nevertheless. Maybe we could have a CI setup using something like https://www.appveyor.com/

cmb69:
We start with a single job.

This is based on work by @weltling.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,7 +7,7 @@ install:
   - ps: appveyor\install.ps1
 
 cache:
-  - c:\build-cache -> .appveyor.yml, appveyor\*
+  - c:\build-cache -> .appveyor.yml, appveyor\install.ps1
 
 environment:
   BIN_SDK_VER: 2.2.0
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -28,6 +28,38 @@ environment:
       ARCH: x86
       TS: 1
       VC: vc15
+    - PHP_VER: 7.3.0
+      ARCH: x64
+      TS: 1
+      VC: vc15
+    - PHP_VER: 7.3.0
+      ARCH: x64
+      TS: 0
+      VC: vc15
+    - PHP_VER: 7.3.0
+      ARCH: x86
+      TS: 1
+      VC: vc15
+    - PHP_VER: 7.3.0
+      ARCH: x86
+      TS: 1
+      VC: vc15
+    - PHP_VER: 7.2.0
+      ARCH: x64
+      TS: 1
+      VC: vc15
+    - PHP_VER: 7.2.0
+      ARCH: x64
+      TS: 0
+      VC: vc15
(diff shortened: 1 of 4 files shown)


cmb69:
Note that this is work in progress; I can't say for sure that packaging actually works, since the build is failing.

lstrojny:
@cmb69 if you rebase against master, the build should be ok now

lstrojny:
Awesome, https://ci.appveyor.com/project/cmb69/php-amqp builds

cmb69:
Rebased, and looks good indeed (https://ci.appveyor.com/project/cmb69/php-amqp/builds/31966330).

The generated packages yet lack rabbitmq.4.dll, what might be okay, though.

Also, I think that actually setting up a build matrix would be useful. Which PHP versions do you like to support?

And, of course, running the test suite would be useful; I wouldn't know how to set that up, though.

lstrojny:
Great!

Knowing nothing about WIndows `rabbitmq.4.dll` would be the relevant shared object of librabbitmq I presume. I don't know what the practice is on Windows with regards to shipping PHP extensions and shared libraries.

For the build matrix: right now we are supporting PHP 5.6 and up, so 5.6, 7.0, 7.1.,7.2, 7.3 and 7.4 is what we build against on Travis CI. I don’t know what is realistic/easy to achieve on Windows, the more the better but generally I think that each version more we targeting is progress.

In order to get the test suite running, RabbitMQ would need to be installed. Here are the docs on how to do it: https://www.rabbitmq.com/install-windows.html. But we can also leave that for another day. Going from no CI for Windows to a single Windows build is an infinite improvement in my book, so thank you already!

lstrojny:
Let me know how you want to proceed here. We can either merge this now, again, it’s already an ∞% improvement or add a few more to the build matrix.

cmb69:
Yes, rabbitmq.4.dll would be the shared object of librabbitmq. This DLL is contained in the PECL packages for Windows, so it makes some sense to include it in the artifacts. I've added a respective commit.

Regarding the build matrix: it shouldn't be a problem to build for 6 PHP versions, but on Windows there are typically 4 variants (x64 vs. x86 and NTS vs. ZTS) for each version, what takes a while to build on AppVeyor (which likely runs the jobs more or less sequentially). So it's mostly about a trade-off between quick feedback from CI vs. running all supported builds. I've added [jobs for all currently built PECL packages](https://ci.appveyor.com/project/cmb69/php-amqp/builds/31970268) in a respective commit. That *might* be a sensible compromise for now.

Regarding running the test suite: I have no experience with RabbitMQ, so I'd like to leave that for another day (or to somebody else). :)

I think this PR is good to be merged. Should probably be squashed; if you want me to do this, just say so. :)

PR #375, opened 2020-04-07
config.w32: add \include\librabbitmq for amqp.h

Jan-E:
Often in Windows build environments the headers are stored in include\{libname}
Add \include\librabbitmq to the paths where we look for amqp.h

### DIFF
--- a/config.w32
+++ b/config.w32
@@ -1,7 +1,7 @@
 ARG_WITH("amqp", "AMQP support", "no");
 
 if (PHP_AMQP != "no") {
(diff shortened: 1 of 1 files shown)


Jan-E:
@cmb69 Is there a compelling reason why the downloads on php.net do not use the include\libname convention? See https://windows.php.net/downloads/pecl/deps/librabbitmq-0.9.0-vc15-x64.zip for instance.

cmb69:
@Jan-E, that mimics a typical include file structure on Linux.

Jan-E:
@cmb69 OK. But I guess there is no objection to facilitate a typical include file structure on Windows as well, like it is done in this PR.

lstrojny:
I am out of my depth here and need opinions of folks.

cmb69:
I think this PR makes some sense. :)

PR #380, opened 2020-07-27
Fix #379: Compile failure on php 8.0

### LINKED ISSUE #379: Compiling fails with PHP 8
I tried to get the extension installed in a php:8.0-rc-cli-alpine3.12 docker image using pickle with the attached log output.

There are PHP 8 breaking changes with removing TSRMLS_CC, TSRMLS_DC macros.

I will try to compile again after removing those macros, ignoring if it would make amqp to fail in ZTS mode

[log.txt](https://github.com/php-amqp/php-amqp/files/4981631/log.txt)

Coffelius:
added a php8_support.h included where it's required in order to get the build working

### DIFF
--- a/amqp_type.h
+++ b/amqp_type.h
@@ -27,7 +27,10 @@
 #include "php.h"
 
 #include <amqp.h>
-#if PHP_MAJOR_VERSION >= 7
+#if PHP_MAJOR_VERSION >= 8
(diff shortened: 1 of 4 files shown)


lstrojny:
Sorry, I didn't get the timing on the PRs right, so I merged the #381 instead. Thank you!

PR #396, opened 2021-04-06
PHP 8 Windows - fix https://github.com/php-amqp/php-amqp/issues/390

Jan-E:
This PR fixes https://github.com/php-amqp/php-amqp/issues/390 by replacing `# include "win32/php_stdint.h"` with `# include "main/php_stdint.h"` for all PHP 8+ versions.

See https://github.com/php-amqp/php-amqp/issues/386 as well.

Please review and merge.

### DIFF
--- a/amqp.c
+++ b/amqp.c
@@ -31,7 +31,11 @@
 #include "zend_exceptions.h"
 
 #ifdef PHP_WIN32
-# include "win32/php_stdint.h"
+# if PHP_VERSION_ID >= 80000
+#  include "main/php_stdint.h"
+# else
+#  include "win32/php_stdint.h"
+# endif
 # include "win32/signal.h"
 #else
 # include <stdint.h>
--- a/amqp_basic_properties.c
+++ b/amqp_basic_properties.c
@@ -36,7 +36,11 @@
 
 #ifdef PHP_WIN32
 # include "win32/unistd.h"
-# include "win32/php_stdint.h"
+# if PHP_VERSION_ID >= 80000
+#  include "main/php_stdint.h"
+# else
+#  include "win32/php_stdint.h"
+# endif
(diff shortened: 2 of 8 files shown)


Jan-E:
Compiled DLL’s for PHP 8 can be found in the zips here:
https://www.apachelounge.com/viewtopic.php?t=6359
The zips also contain a rabbitmq.4.dll, built from the most recent rabbitmq-c sources yesterday.

cmb69:
Did you try to not include either header? IIRC, that should be fine for VC14 and higher.

Jan-E:
> Did you try to not include either header? IIRC, that should be fine for VC14 and higher.

No, I did not. I assumed the include php_stdint.h was there for a purpose. Note that there is not a `#include <stdint.h>` in the code (for Windows builds). Should not one of the 2 be there?

cmb69:
Ah, right! I think just including `<stdint.h>` on Windows should be fine.

Jan-E:
I ran the build for PHP 8.0.8 on Appveyor:
https://ci.appveyor.com/project/Jan-E/php-amqp/history

@cmb69 Could you add vs16 versions for rabbitmq at
https://windows.php.net/downloads/pecl/deps/

Fot this appveyor test I used the vc15 versions.

cmb69:
> Could you add vs16 versions for rabbitmq at
https://windows.php.net/downloads/pecl/deps/

Done (just copied and renamed; no time to actually do an update; sorry!)

Jan-E:
Thanks. Works: https://ci.appveyor.com/project/Jan-E/php-amqp/history

Let me know when you have real VS16 deps. In my own builds I am using librabbitmq 0.12.0-pre at the moment. A bit newer than 0.9.0.

PR #414, opened 2022-04-03
Fix ack/nack/reject param documentation

flavioheleno:
The stubs document **ack**, **nack** and **reject** methods with `$delivery_tag` as being `string` instead of `integer` as can be seen in:

* https://github.com/php-amqp/php-amqp/blob/master/amqp_queue.c#L760
* https://github.com/php-amqp/php-amqp/blob/master/amqp_queue.c#L805
* https://github.com/php-amqp/php-amqp/blob/master/amqp_queue.c#L851

This PR updates it accordingly.


lyrixx (approved):

lstrojny:
Thank you very much!

PR #415, opened 2022-04-03
Fix AMQPEnvelope::getDeliveryTag() return type

flavioheleno:
The stub for **AMQPEnvelope** defines `getDeliveryTag()` as returning `string` instead of `integer` as can be seen in:

* https://github.com/php-amqp/php-amqp/blob/master/amqp_envelope.c#L77

This PR updates it accordingly.


lyrixx (approved):

lstrojny:
Thank you!

PR #435, opened 2023-07-26
More consistent return types for AMQPEnvelope

lstrojny:
* `AMQPEnvelope::getBody()` always returns a string now while previously it returned false for an empty body
 * `AMQPEnvelope::getHeader()` returns null for an unknown header instead of false
 * `AMQPEnvelope::getRoutingKey()` always returns a string now

### DIFF
--- a/stubs/AMQPEnvelope.php
+++ b/stubs/AMQPEnvelope.php
@@ -30,7 +30,7 @@ class AMQPEnvelope extends AMQPBasicProperties
     /**
      * Get the consumer tag of the message.
      *
-     * @return string The consumer tag of the message.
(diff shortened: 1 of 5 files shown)


stubs/AMQPEnvelope.php
```diff
@@ -30,7 +30,7 @@ public function getRoutingKey()
     /**
      * Get the consumer tag of the message.
      *
-     * @return string The consumer tag of the message.
+     * @return string|null The consumer tag of the message.
```
Tobion, line 33: you could switch to real type declarations, can't you?

lstrojny: Yup, that's true

PR #486, opened 2023-09-07
Remove assert on undefined variable

remicollet:


enumag:
I ran into an error while using `pecl install amqp`:

```
/tmp/pear/temp/amqp/amqp_connection_resource.c: In function 'php_amqp_connection_resource_error_advanced':
/tmp/pear/temp/amqp/amqp_connection_resource.c:239:12: error: 'resource' undeclared (first use in this function)
  239 |     assert(resource != NULL);
```

Which looks like it should be fixed by this patch. Can you release a new version with this fix?

remicollet:
@enumag this error is raised when build against debug version of PHP
In such case, there is much more to fix, e.g. see #490

enumag:
Yeah it was against a debug build. I ran into a segfault this morning so I'm trying to build a new debug image with all of the extensions I need to debug it.

Okay I'll keep an eye on the PR you linked and use the 1.x version of amqp for now.

remicollet:
BTW, Why are you using a debug build ?

Debug builds are designed for PHP and extensions developers, not for other users.

enumag:
@remicollet Because I need the [gdb backtrace](https://bugs.php.net/bugs-generating-backtrace.php) of the segfault in order to report it to PHP team.

remicollet:
> @remicollet Because I need the [gdb backtrace](https://bugs.php.net/bugs-generating-backtrace.php) of the segfault in order to report it to PHP team.

Bad choice ;)

You can use GDB and generate backtrace with standard builds.
The documentation seems wrong and outdated.

enumag:
Oh... Well the last time I tried it I was getting question marks in the gdb backtrace instead of filenames. If that's not caused by not using debug mode, does that mean it's caused by this line in the php docker image?

https://github.com/docker-library/php/blob/e9e47fd185d6b36e491c713811437b0c17944faa/Dockerfile-linux.template#L400

remicollet:
Yes `strip` removes all debug symbols from binary to reduce their size.

In linux distro, you usually have debug symbols available separately
(ex `debuginfo` and `debugsource` for RPM, gdb is even able to install them automatically when invoked).

enumag:
Good to know. Do you have an idea how significant is the size reduction? Is it worth it?

remicollet:
> Good to know. Do you have an idea how significant is the size reduction? Is it worth it?

Yes, worth it.

and FYI, documentation is fixed

PR #571, opened 2024-06-21
feat(exchange): treat various ssl and tcp errors as `AMQPConnectionException`

### LINKED ISSUE #48241: [Messenger] Issues after migrating to AWS RabbitMQ
### Symfony version(s) affected

6.0

### Description

We are switching from normal RabbitMQ server, to [AWS-managed cluster](https://docs.aws.amazon.com/amazon-mq/latest/developer-guide/working-with-rabbitmq.html).

After that periodically we have this error in log:
```
(AMQPException): Library error: a SSL error occurred
```
It happens on line `$exchange->publish(...)` in `\Symfony\Component\Messenger\Bridge\Amqp\Transport\Connection::publishOnExchange`

The problem is that after some time, the connection becomes "dead", but method `Connection::clearWhenDisconnected` doesn't detect is as disconnected.

It happened after 350 seconds, because they have some [idle timeout](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#connection-idle-timeout).

I've provided a small script, that reproduces the issue: https://gist.github.com/xorik/24065aab0db094c99e746b277466fcba

Here's output:
```
Channel is connected!
Published!
Now wait for 360 seconds....
Channel is connected!
Exception (AMQPException): Library error: a SSL error occurred
#0 /tmp/1.php(37): AMQPExchange->publish('howdy-do', 'hello')
#1 {main}
```

I can give you credentials to a test server, if you want (PM https://twitter.com/xorik_dev/ or give me your contact).

I've tried to make persistent connection, adding heartbeat, but it won't help.

This is probably connected with https://github.com/symfony/symfony/issues/36538, but suggested fix doesn't help.

### How to reproduce

- Connect to a AWS RabbitMQ
- publish a message
- wait 350+ seconds
- publish a message

### Possible Solution

- Actually the connection isn't 100% dead, so reconnect is not required and this code actually fixes the issue:
 
```php
try {
    $exchange->publish($message, $routing_key);
} catch (\AMQPException) {
    $exchange->publish($message, $routing_key);
}
```

- Alternative solution is perform some kind of ping or heartbeat, before publishing a message (improve method `clearWhenDisconnected`)

### Additional Context

This is clearly not a Symfony issue, but I think Messenger should handle this issue, since this AWS RabbitMQ is quite popular, and there is no easy way to workaround it by developers, except creating a new transport and copy+pasting or extending bunch of Symfony classes.

tunterreitmeier:
This tries to fix symfony/symfony/issues/48241

The specific issue is that e.g. the AWS Load Balancer closes idle connections after a certain time. Long running tasks will then run into the issue of a lost connection.
symfony/symfony/pull/54167 already provides an adequate solution by simply trying to reconnect. However, the fix specifically catches `AMQPConnectionException`, while the actual exception thrown is the less specific `AMQPException`.
This can also be seen in the original issue:

> (**AMQPException**): Library error: a SSL error occurred

My first instinct was to catch the less specific Exception instead, which would solve the issue.
https://github.com/symfony/amqp-messenger/compare/6.4...wikando:amqp-messenger:6.4-retry
But I think there is a fair argument to be made, that this indeed is an Exception relating to the connection. That's where this PR is coming from.
I tried coming up with a test, but I'm at a loss how to recreate this specific scenario (connection closed by server). Feedback is obviously welcome.

I really hope that helps. Thanks a bunch!

### DIFF
--- a/amqp.c
+++ b/amqp.c
@@ -491,12 +491,23 @@ void php_amqp_zend_throw_exception(
             break;
         case AMQP_RESPONSE_LIBRARY_EXCEPTION:
             switch (reply.library_error) {
+                case AMQP_STATUS_BROKER_UNSUPPORTED_SASL_METHOD:
                 case AMQP_STATUS_CONNECTION_CLOSED:
+                case AMQP_STATUS_HOSTNAME_RESOLUTION_FAILED:
                 case AMQP_STATUS_SOCKET_ERROR:
                 case AMQP_STATUS_SOCKET_CLOSED:
                 case AMQP_STATUS_SOCKET_INUSE:
-                case AMQP_STATUS_BROKER_UNSUPPORTED_SASL_METHOD:
-                case AMQP_STATUS_HOSTNAME_RESOLUTION_FAILED:
+                case AMQP_STATUS_SSL_CONNECTION_FAILED:
+                case AMQP_STATUS_SSL_ERROR:
+                case AMQP_STATUS_SSL_HOSTNAME_VERIFY_FAILED:
+                case AMQP_STATUS_SSL_PEER_VERIFY_FAILED:
+#if AMQP_VERSION >= AMQP_VERSION_CODE(0, 11, 0, 1)
+                case AMQP_STATUS_SSL_SET_ENGINE_FAILED:
+#endif
+#if AMQP_VERSION >= AMQP_VERSION_CODE(0, 14, 0, 1)
+                case AMQP_STATUS_SSL_UNIMPLEMENTED:
+#endif
+                case AMQP_STATUS_TCP_ERROR:
                     exception_ce = amqp_connection_exception_class_entry;
                     break;
                 default:


lstrojny:
Could you rebase?

tunterreitmeier:
done - thanks for looking into this

lstrojny:
Looks like we have failures for librabbitmq <= 0.13. You probably need to add a preprocessor conditional based on `AMQP_VERSION`

tunterreitmeier:
It's been a while since I checked librabbitmq, but yeah two consts are not defined in older versions.
Was able to build and test with v.010.0

lstrojny:
I would have a relatively strong preference of not checking for the literal constants but instead for the AMQP_VERSION (see https://github.com/alanxz/rabbitmq-c/blob/8b7471eab8d09536b3c104dbb30a65699cf48104/include/rabbitmq-c/amqp.h#L157 and packing macro in https://github.com/alanxz/rabbitmq-c/blob/8b7471eab8d09536b3c104dbb30a65699cf48104/include/rabbitmq-c/amqp.h#L136). That way, when we remove support for older librabbitmq versions it is very clear which conditions can be retired. Because I know that otherwise these ifdefs will linger around forever – because I will totally forget.

tunterreitmeier:
Totally makes sense. The check now uses `AMQP_VERSION`

tunterreitmeier:
@lstrojny can you retry the single failing pipeline? I feel like it is just some flakyness in the setup-php action

PR #584, opened 2024-09-19
Updated composer.json to enable support for PIE

asgrim:
Adds support for [PIE](https://github.com/php/pie).

Once merged, a maintainer would need to add this repo's URL to https://packagist.org/packages/submit :)

More info/documentation can be read on: https://github.com/php/pie/blob/main/docs/extension-maintainers.md

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
   "name": "php-amqp/php-amqp",
-  "type": "library",
+  "type": "php-ext",
   "description": "PHP AMQP Binding Library",
   "keywords": [
     "rabbitmq",
@@ -31,5 +31,19 @@
     "allow-plugins": {
       "dealerdirect/phpcodesniffer-composer-installer": false
     }
+  },
+  "php-ext": {
+    "extension-name": "amqp",
+    "configure-options": [
+      {
+        "name": "with-amqp",
+        "description": "Include amqp support"
+      },
+      {
+        "name": "with-librabbitmq-dir",
+        "description": "Set the path to librabbitmq install prefix.",
+        "needs-value": true
+      }
+    ]
   }
 }


composer.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "php-amqp/php-amqp",
```
asgrim, line 2: Please verify vendor namespace & package name are acceptable

ruudk:
@lstrojny What do you think of this? It would be great to be able to install AMQP using PIE.

composer.json
pinepain: Good point. The ext name itself is `amqp`, see https://github.com/php-amqp/php-amqp/blob/a9cbd81ea401d4ba41b72381008b331f5516823c/amqp.c#L386 so given the constrain on ext name in PIE tool itself (https://github.com/php/pie/blob/c6f0f4b2bfcc1de7e5f55b6317ef4d36b2ce49f2/src/ExtensionName.php#L26-L31), should it be named as `php-amqp/amqp` then ? And does PIE impose limitations on what vendor name could be?

asgrim: The *Package Name* does not have any constraints. But the *Extension Name* always has done. 

I have defined the *Extension Name* according to the existing `package.xml` as `amqp` on L36, so the *Package Name* can really be anything that makes sense for the project; same with the vendor 👍

pinepain: Got you, thanks for clarification.

*(might be unrelated to the scope of PR work)* Interesting enough, the composer package is still pointing to the old repo name - `pdezwart/php-amqp`. And the reason we got composer.json is to be able to fetch stub files locally, it was the days before PHPStorm had it bundled and I guess the PHPStorm itself was not that widely adopted itself, afair

asgrim: https://packagist.org/packages/php-amqp/php-amqp has not yet been pushed to Packagist, so I don't think this should be an issue? :thinking:

pinepain:
@asgrim this extension depends on librabbitmq been present and of a specific version, I've looked into the PIE docs (https://github.com/php/pie/blob/main/docs/extension-maintainers.md#extension-dependencies) and the way I read it is that currently there is no (reasonable) way do declare such dependencies with PIE, do I understand it correctly?

asgrim:
> @asgrim this extension depends on librabbitmq been present and of a specific version, I've looked into the PIE docs (https://github.com/php/pie/blob/main/docs/extension-maintainers.md#extension-dependencies) and the way I read it is that currently there is no (reasonable) way do declare such dependencies with PIE, do I understand it correctly?

That is correct, that much has not (yet) changed from PECL. See also: php/pie#109

composer.json
pinepain: Not an issue at all, we would just need to set up a redirect from that old one to a new one, afair there is sth like that in packagist. We would likely need Lars or Pieter for that though.

pinepain:
Thanks for such a quick answer. Indeed it's a challenging task. I'm impressed to see PIE, actually. I was thinking about sth like that back in days I was spending time on this and other extensions, but ended with the conclusion that `pecl` does the job just fine. Though, it should be quite an improvement for end-user to install with PIE the way it looks like. Would be practical to ask here over PIE vs pecl or was it already discussed in PIE itself? If it was already discussed, I would appreciate a link, curious to learn about PIE and it's roadmap.

asgrim:
Sure - the design doc is at https://github.com/ThePHPF/pie-design

The longer term goal is to replace PECL completely and retire pecl.php.net, but we need to reach some maturity first :)

pinepain:
That's an impressive bunch of work you've already done, thank you and other contributors of PIE!

And pardon for "hijacking" the discussion here.

The PR lgtm and should bring no maintenance burden. After the merge, one with the proper access to packagist would have to upload the new package and, ideally, mark the old one as abandoned and point to the new name.

ruudk:
@pinepain does that mean we can merge this? I would really like to be able to install this great extension through 🥧

pinepain:
Technically yes, but I don't have merge privilege any longer (I've been away for a while), so I guess the only hope is Lars will be nearby or that there is another maintainer to have it merged and tag a release or so.

lstrojny:
@pinepain excited to see you again!

Excited about PIE as well, will hopefully make C-extensions much more accessible.

ruudk:
Thanks for the merge. There might be something still pending as it doesn't show up yet: https://packagist.org/extensions?query=php-amqp%2Fphp-amqp

pinepain:
I think the issue is this:

> After the merge, one with the proper access to packagist would have to upload the new package and, ideally, mark the old one as abandoned and point to the new name.

asgrim:
@lstrojny any chance of getting this added to Packagist? :)

asgrim:
FYI @lstrojny I added https://packagist.org/packages/php-amqp/php-amqp to Packagist and you as a maintainer; it is not auto-updated (would need the hook setup for that to work), but at least it is there now :heavy_check_mark:

ghost:
👋 when I try to install php-amqp using pie `pie install php-amqp/php-amqp`

```
0.223 🥧 PHP Installer for Extensions (PIE) 1.0.0, from The PHP Foundation                                                                                                            
0.390 You are running PHP 8.4.7                                                                                                                                                       
0.404 Target PHP installation: 8.4.7 nts, on Linux/OSX/etc x86_64 (from /usr/local/bin/php)                                                                                           
0.861 
0.863 In UnableToResolveRequirement.php line 36:
0.863                                                                                
0.863   Package php-amqp/php-amqp was not of type php-ext or php-ext-zend (requeste  
0.863   d php-amqp/php-amqp).                                                        
0.863                                                                                
0.863 
0.863 install [-j|--make-parallel-jobs MAKE-PARALLEL-JOBS] [--skip-enable-extension] [--force] [-d|--working-dir WORKING-DIR] [--with-php-config WITH-PHP-CONFIG] [--with-php-path WITH-PHP-PATH] [--with-phpize-path WITH-PHPIZE-PATH] [--] [<requested-package-and-version>]
0.863 
```

Maybe a question of time until the extension to be available on packagist?

asgrim:
@qdequippe-pdp Yes, that is slightly confusing; that is because older stable tagged versions containing the `composer.json` were not PIE packages at the time.

Until a new stable tag is made, using a branch alias should work, such as:

```
pie install php-amqp/php-amqp:dev-latest
```

ruudk:
Is it possible to tag a new release so that it works immediately?? Thanks 🙏🏻

PR #593, opened 2025-03-21
Fix clang format errors

jeremy-smith-maco:
Hoping to fix any issues so that a release can be done. :)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -30,7 +30,7 @@ jobs:
           concurrent_skipping: 'same_content_newer'
 
       - name: Checkout
-        uses: actions/checkout@v4.1.1
+        uses: actions/checkout@v4.2.2
 
       - name: Prepare matrix
         id: matrix
@@ -53,7 +53,7 @@ jobs:
           install: clang-format-17
 
       - name: Checkout
-        uses: actions/checkout@v4.1.1
+        uses: actions/checkout@v4.2.2
 
       - name: Check style
         run: ./infra/tools/pamqp-format-check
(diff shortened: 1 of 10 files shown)


jeremy-smith-maco:
Not sure what these errors are caused by

> Error: Missing download info for actions/upload-artifact@v3

Edit: Looks like it needs to upgrade to v4 as of January 30th

Edit: Fix that error but some tests seem to be failing still

jeremy-smith-maco:
I've made some progress on the failing tests.

- At some point the headers got changed to being sorted by the key - couldn't find where this happened
- The invalid exchange type error was changed to a precondition failed error instead and doesn't kill the connection (https://github.com/rabbitmq/rabbitmq-server/pull/11214)
- The `x-death` header is not interpreted by rabbitmq 4.0+ anymore so had to change it to something else - I used `x-death-new` (https://github.com/rabbitmq/rabbitmq-server/releases/tag/v4.0.1)

Now seems to be an issue with cmake when building librabbitmq, but it isn't consistent. Sometimes fails, sometimes not.

jeremy-smith-maco:
Looks like the runner, both ubuntu 20.04 and 22.04, sometimes uses a slightly newer version that has cmake 4 in it which breaks and sometimes uses a slightly older version which is fine so it randomly works.

Looks like they are completely removing the 20.04 runner very soon as well so something will have to be done here to fix it. 

https://github.com/actions/runner-images/issues/11101

jeremy-smith-maco:
These changes should make the tests consistently pass. I have done the following, but this can be changed if there is a more preferred approach of course:

* Manually install clang 11 and cmake 3.22.6 (both are cached so should realistically only have to do it once) - this avoids having to use cmake 4 randomly which breaks the build
* If testing rabbitmq 0.10.0 or below, will manually install openSSL 1.1.1 and force cmake to use this when building librabbitmq (also cached, so should only be installed once)
* Always uses ubuntu 22.04 since 20.04 will be gone soon. Didn't test with 24.04 but may also just work with that

If it all looks good, tests pass and you're happy with the changes, please merge and release if possible. :)

jeremy-smith-maco:
@lstrojny Just tagging you in-case you haven't seen this please :)

jeremy-smith-maco:
@bkw Any chance you are able to take a look at this and maybe release a new version if all is good please?

dzuelke:
@lstrojny ping - could you merge this PR? This would unblock a bunch of other stuff that currently has CI failing, e.g. #595

lstrojny:
Has been resolved in https://github.com/php-amqp/php-amqp/commit/8b76bad74a7954839d8808130d6cc9db435532ac

PR #597, opened 2025-10-23
Add PHP 8.4 and 8.5 to CI matrix

ruudk:


ruudk:
@lstrojny Hi Lars, could you please approve this workflow so that we can see what the results are? Thanks 🙏

PR #599, opened 2025-11-13
Fix segmentation fault in php_amqp_close_channel #598

### LINKED ISSUE #598: Segmentation fault when channel closes
Following on from https://github.com/php-amqp/php-amqp/issues/588

There is a use-after-free bug in php_amqp_close_channel

If amqp_channel_close errors with PHP_AMQP_RESOURCE_RESPONSE_ERROR_CONNECTION_CLOSED or PHP_AMQP_RESOURCE_RESPONSE_ERROR if forces the connection to disconnect.
php_amqp_disconnect_force calls into php_amqp_cleanup_connection_resource which calls zend_list_delete(resource) and frees up the connection_resource object when it calls amqp_connection_resource_dtor/connection_resource_destructor.
Image
If this happens inside php_amqp_close_channel it then goes on to mark the slot as used, but connection_resource may have already been freed by the php_amqp_disconnect_force  call.
Image

pauljohnston2009:
Closes https://github.com/php-amqp/php-amqp/issues/598


lstrojny:
Could you rebase against latest branch so that we get a clean build?

PR #619, opened 2026-04-30
Replace XtOffsetOf() with offsetof() for PHP 8.6

andypost:
PHP 8.6 removed XtOffsetOf(), which was an alias for the standard C offsetof() macro. Replace all usages with offsetof() directly.

Ref: https://github.com/php/php-src/commit/7114314c5a96c362b95663f7e7c9184586721f58


remicollet:
@andypost sorry to have missed this PR

BTW see PR #638

PR #627, opened 2026-06-04
feature: non-blocking consume added

adapik:
# Add `AMQP_NB_CONSUME` flag for non-blocking, no-throw consuming

Solves: (https://github.com/php-amqp/php-amqp/issues/628)

## Summary

Adds an `AMQP_NB_CONSUME` flag that lets you drain an already-established consumer **non-blockingly**: if there is a frame buffered locally it is delivered to the callback, and if the buffer is empty `consume()` returns control to PHP immediately **without** throwing `Consumer timeout exceed`.

## Motivation

Today the consume loop has only two waiting modes:

- `read_timeout == 0` → block indefinitely;
- `read_timeout > 0` → wait, then throw `AMQPQueueException("Consumer timeout exceed")`.

There is no way to treat "buffer is empty right now" as a normal control-flow signal. Building a self-driven main loop therefore means catching an exception on every empty poll, which is both awkward and wasteful. `AMQP_NB_CONSUME` fills exactly that gap.

## What changed

- **`php_amqp.h`** — define `AMQP_NB_CONSUME` and register the constant.
- **`amqp_queue.c`** — when the flag is set, force a zero `timeval` for the internal wait, and in the `AMQP_STATUS_TIMEOUT` branch return cleanly instead of throwing. All other library-exception / connection-error paths are untouched.
- **`stubs/AMQP.php`, `stubs/AMQPQueue.php`** — document the new flag.
- **`tests/`** — new `.phpt` cases (see below).

## Semantics

- With `AMQP_NB_CONSUME` the internal wait uses a zero `timeval`, so there is no socket blocking.
- A buffered frame is delivered to the callback as usual, with no socket round-trip.
- An empty buffer returns control to PHP silently — no exception.
- Only `AMQP_STATUS_TIMEOUT` is swallowed; genuine connection/library errors are surfaced exactly as before.
- QoS is respected: if prefetch left several messages buffered, returning `false` from the callback after taking one leaves the rest for the next call.

Intended usage, combined with the existing `AMQP_JUST_CONSUME`:

```php
$queue->consume(null); // subscribe once

while ($running) {
    $msg = null;
    $queue->consume(function (AMQPEnvelope $e) use (&$msg) {
        $msg = $e;
        return false;
    }, AMQP_JUST_CONSUME | AMQP_NB_CONSUME);

    if ($msg !== null) {
        // handle + ack
    }
    // other work / signals / limits
}
```

## Backward compatibility

Fully additive. Behavior is identical to before unless `AMQP_NB_CONSUME` is explicitly passed. No existing flag values change.

## Implementation notes

The non-blocking wait uses librabbitmq's own zero-`timeval` path rather than a raw `select()`, so it does not break the library abstraction or introduce portability issues, and partial-frame reassembly is still handled internally by librabbitmq.

## Testing

Added `.phpt` cases modeled on `amqpqueue_consume_timeout.phpt`:

- empty buffer with `AMQP_JUST_CONSUME | AMQP_NB_CONSUME` returns immediately and throws nothing (with timing assertion to prove it does not block);
- a previously published message is delivered to the callback in non-blocking mode.


nicolas-grekas (approved):
I needed exactly this while adding push consumption to Symfony's AMQP transport (symfony/symfony#65924), so I built this branch against PHP 8.4 and put it through the cases I care about. It does what it says:

```
1. empty buffer, JUST_CONSUME|NB_CONSUME:  returned in 0.0002s, no exception
   same call without the flag:             1.0003s, AMQPQueueException "Consumer timeout exceed"
2. 8 messages already pushed:              all 8 delivered in 0.0004s
3. 100 bodies of 10B to 700KB (many
   frames each), drained through NB:       intact and in order
4. after disconnect():                     AMQPChannelException still raised
```

Point 4 matters: only `AMQP_STATUS_TIMEOUT` is swallowed, so a genuine failure is still surfaced. And 32768 is the next free bit after `AMQP_REQUEUE`, with no overlap with the flags `basic.consume` itself reads (`AMQP_NOLOCAL`, `AMQP_AUTOACK`, `AMQP_EXCLUSIVE`).

### Why this is worth having

Without it, "the buffer is empty" can only be learned by catching an exception, and that turns out to cost more than the ugliness. Until php/php-src#23624, the engine drops every signal queued while an internal function was running when that function returns with an exception pending: `ZEND_VM_FCALL_INTERRUPT_CHECK()` sits right after an internal call returns, before the exception is handled, so `pcntl_signal_dispatch()` runs with `EG(exception)` set, `zend_call_function()` no-ops, and the queued signals are recycled without a single handler running.

A worker looping on `consume()` spends essentially all of its wall time inside that call, so it misses **every** `SIGTERM`:

```
read_timeout=0.20s, SIGTERM at 1.70s: 31 rounds, never stopped
read_timeout=0.05s, SIGTERM at 2.13s: 121 rounds, never stopped
read_timeout=0.01s, SIGTERM at 2.57s: 653 rounds, never stopped
```

653 iterations of a 10 ms read timeout and the signal never arrives, not even through an explicit `pcntl_signal_dispatch()` afterwards. Not blocked and not ignored: `SigBlk`, `SigIgn` and `SigCgt` read from inside the consume callback confirm the handler is installed, and the same signal during `sleep()` or `stream_select()` is delivered normally. What makes `consume()` different is only that it throws, which is why a no-throw drain is not just cosmetic.

### One doc suggestion

`AMQP_NB_CONSUME` is non-blocking with respect to the *next* message, not mid-message: once the method frame is read, `amqp_read_message()` waits for the header and body frames with no timeout. That is the right behaviour, and it is what keeps the 700KB bodies in my test 3 intact, but the stub currently reads as an unqualified "returns immediately", which could surprise someone on a slow link. Something like "returns as soon as no further message is buffered; a message already being received is still read to completion" would set the expectation better.

Worth noting too that with a callback returning `true` the flag does not mean "drain what is buffered right now and stop": it keeps going while frames keep arriving. All 100 messages in my test 3 came back in a single call. That is fine, just not obvious from the name.

Nothing blocking from me. Happy to see this land, it would let the Symfony bridge drop a `setReadTimeout(0.001)` workaround.

adapik:
Thanks a lot for the thorough review, @nicolas-grekas! Really appreciate you putting the branch through such a rigorous set of tests, including the signal-handling investigation — that's a great catch and explains a real-world pain point.

I looked into you work [here](https://github.com/symfony/symfony/pull/65924). I had the same intention behind this PR: I actively use Symfony's AMQP Messenger transport myself and wanted to improve its performance, and I can see (and value) how much work you've already put into this, both in testing this PR and in tracking down the SIGTERM/exception-swallowing bug.

I've updated the documentation in the stubs based on your comments — it should now be more precise about the "non-blocking with respect to the next message" semantics and the fact that a callback returning true will keep draining as long as frames keep arriving.

@lstrojny, would you mind taking a look when you have a moment? Given @nicolas-grekas's comment above, this seems to have real, concrete value for a very large consumer of this extension (Symfony), so a review would be much appreciated. Ping me if any changes required.

PR #638, opened 2026-07-16
fix build with PHP 8.6.0alpha2

remicollet:
from UPGRADING.INTERNALS

  . The XtOffsetOf() alias of C’s offsetof() macro has been removed. Use
    offsetof() directly.

  . The INI_STR(), INI_INT(), INI_FLT(), and INI_BOOL() macros have been
    removed. Instead new zend_ini_{bool|long|double|str|string}_literal()
    macros have been added.

Also add the new macros for compatibility with old PHP versions


lstrojny:
Looks great, thank you. Can you check the formatting error?

remicollet:
> Looks great, thank you. Can you check the formatting error?

Sorry, I have no idea what `"code should be clang-formatted"` means (and doesn't seems related to this change)

